PR #4, opened 2022-03-31
bug 577511 - do not ignore cancel

jukzi:
Ignoring cancel leads to loss of changes

Change-Id: I02c57918c4d280794386a3c78b2ff566c4051259

### DIFF
--- a/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
+++ b/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
@@ -89,6 +89,7 @@ import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.SafeRunner;
 import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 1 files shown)


kthoms (commented):
LGTM

org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java
```diff
@@ -4896,6 +4897,10 @@ protected void validateState(IEditorInput input) {
 
 		} catch (CoreException x) {
 			IStatus status= x.getStatus();
+			if (status != null && status.getSeverity() == IStatus.CANCEL) {
```
laeubi, line 4900: Maybe an IStatus.checkCancel() throws OperationCanceledException would be useful instead of coding checks explicitly.

jukzi: i see a lot getSeverity() == IStatus.CANCEL in the codebase but with very various implementations.

laeubi: One more reason for me to unify this.

jukzi: feel free to make another ticket for that. but i do not see a way how todo that.

iloveeclipse:
@jukzi : I don't see **any** effect from this change **except** additionally to the original problem  we have now this: https://github.com/eclipse-platform/eclipse.platform.text/issues/65.

Performing steps from bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=577511 I see that 
1) there are still multiple "cancel" clicks required 
2) From entered text only few characters are added

Since the original issue is not fixed, and new error is logged, I would like to revert this change. Objections?

jukzi:
Please hold on. i will first try to reproduce - the fix did work for me, so let's see why it does not work anymore.

PR #8, opened 2022-04-06
Refactoring: Convert QuickTextSearchRequestor to interface

kthoms:
Change the type of QuickTextSearchRequestor from class to interface.
The nature of it is more define an interface. The intention of having
a class here might have been to have empty default methods that are
overridden by sub-types. The class was not ever instantiated directly.

### DIFF
--- a/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearchRequestor.java
+++ b/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearchRequestor.java
@@ -13,7 +13,7 @@
 package org.eclipse.text.quicksearch.internal.core;
 
 /**
- * Plays a similar role than SearchReqeustor in eclipse Searches. I.e. a search requestor
+ * Plays a similar role than SearchReqeustor in Eclipse searches. I.e. a search requestor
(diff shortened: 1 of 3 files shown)


org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearchRequestor.java
```diff
@@ -24,30 +24,31 @@
  *
  * @author Kris De Volder
  */
-public class QuickTextSearchRequestor {
+@SuppressWarnings("unused")
+public interface QuickTextSearchRequestor {
```
iloveeclipse, line 28: Shouldn't this be **renamed** too following Eclipse conventions to `IQuickTextSearchRequestor` ?

iloveeclipse (changes requested):

org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/QuickTextSearchRequestor.java
akurtakov: @kthoms Would you please adjust according to @iloveeclipse request?

vogella:
@kthoms do you still plan to update this one?

PR #16, opened 2022-05-04
Example usage of display executor

laeubi:
Example for https://github.com/eclipse-platform/eclipse.platform.swt/pull/67


mickaelistria:
Let's reopen, I'd like to see it merge if it passes with current API.

mickaelistria:
@laeubi I'd like to merge it; but before I do so, I wonder whether it would be better to showcase the future construst insted:
```java
CompletableFuture.runAsync(() -> {
   ....
}, text.getDisplay());
```
What do you think?

laeubi:
`CompletableFuture` mostly makes sense if you want to chain the result, or do certain actions on failure/completion, I'm not sure if this applies here.

PR #22, opened 2022-05-06
AnnotationModel uses inefficient iteration over map (Fixes #21)

### LINKED ISSUE #21: AnnotationModel uses inefficient iteration over map
`AnnotationModel.cleanup(boolean, boolean)` iterates over map with keyset iterator and does lookup for every value from iterator. With ~10_000 annotations this is not fun anymore. Similar code in other places. That can be improved by using entryset iterator.

I plan to push the fix after #19 is done.

iloveeclipse:
AnnotationModel.cleanup(boolean, boolean) iterates over map with keyset
iterator and does lookup for every value from iterator. With ~10_000
annotations this is not fun anymore. Similar code in other places. That
can be improved by using entryset iterator.

See https://github.com/eclipse-platform/eclipse.platform.text/issues/21


org.eclipse.text/src/org/eclipse/jface/text/source/AnnotationModel.java
```diff
@@ -640,13 +640,13 @@ private void cleanup(boolean fireModelChanged, boolean forkNotification) {
 			fDocumentChanged= false;
 
 			ArrayList<Annotation> deleted= new ArrayList<>();
-			Iterator<Annotation> e= getAnnotationMap().keySetIterator();
 			IAnnotationMap annotations= getAnnotationMap();
-			while (e.hasNext()) {
-				Annotation a= e.next();
-				Position p= annotations.get(a);
+			Iterator<Entry<Annotation, Position>> iterator = annotations.entrySet().iterator();
```
szarnekow, line 644: The entrySet is not based on a defensive copy (explicitly according the the JavaDoc of IAnnotationMap), whereas the keysetIterator was save to use. I feel we are now iterating without any locks and a prone to concurrent modifications on the map.

szarnekow: All iterators were creating a copy of the underlying collections which in fact complies with the interface.
Should we get rid of IAnnotationMap.... in theory, yes.
In another context I rewrote the better part of AnnotationModel and friends, but that's not strictly backwards compatible. Not sure really how to deal with it.

szarnekow: What I had in place for quite a few releases was:
```


import java.util.ArrayList;
import java.util.Collection;
import java.util.HashMap;
import java.util.IdentityHashMap;
import java.util.Iterator;
import java.util.Map;
import java.util.Set;
import java.util.function.BiConsumer;

import org.eclipse.jface.text.Position;
import org.eclipse.jface.text.source.Annotation;
import org.eclipse.jface.text.source.IAnnotationMap;

import com.google.common.annotations.VisibleForTesting;

/**
 * Adapted from {@code org.eclipse.jface.text.source.AnnotationMap}
 *
 * Notable changes:
 * - accessible from the current package.
 * - allow to acquire the lock without branching.
 * - maintain the positionToAnnotation map, too.
 */
final class BidirectionalAnnotationMap implements IAnnotationMap {

	/**
	 * The lock object used to synchronize the operations explicitly defined by
	 * <code>IAnnotationMap</code>
	 */
	private Object fLockObject;
	/**
	 * The internal lock object used if <code>fLockObject</code> is <code>null</code>.
	 *
	 * @since 3.2
	 */
	private final Object fInternalLockObject;

	/*
	 * Implementation note:
	 * It would be great if we could use Guava's BiMap instead but that one cannot
	 * be parameterized with a hashing strategy based on Identity. We'd need to wrap
	 * each Object in an Equivalence.Wrapper which would be cumbersome to work with.
	 *
	 * Therefore we resort to two fields for the forward and inverse mapping.
	 */

	/** The map holding the annotations */
	private final Map<Annotation, Position> fInternalMap;

	/** The map holding the positions */
	private final IdentityHashMap<Position, Annotation> fPositions;

	/**
	 * Creates a new annotation map with the given capacity.
	 *
	 * @param capacity the capacity
	 */
	public BidirectionalAnnotationMap(int capacity) {
		fInternalMap= new HashMap<>(capacity);
		fPositions = new IdentityHashMap<>(capacity);
		fLockObject = fInternalLockObject = new Object();
	}

	@Override
	public synchronized void setLockObject(Object lockObject) {
		fLockObject = lockObject != null ? lockObject : fInternalLockObject;
	}

	@Override
	public synchronized Object getLockObject() {
		return fLockObject;
	}

	@Override
	public Iterator<Position> valuesIterator() {
		synchronized (fLockObject) {
			return new ArrayList<>(fInternalMap.values()).iterator();
		}
	}

	@Override
	public Iterator<Annotation> keySetIterator() {
		synchronized (fLockObject) {
			return new ArrayList<>(fInternalMap.keySet()).iterator();
		}
	}

	@Override
	public void forEach(BiConsumer<? super Annotation, ? super Position> action) {
		synchronized (fLockObject) {
			fInternalMap.forEach(action);
		}
	}

	@Override
	public boolean containsKey(Object annotation) {
		synchronized (fLockObject) {
			return fInternalMap.containsKey(annotation);
		}
	}

	@Override
	public Position put(Annotation annotation, Position position) {
		synchronized (fLockObject) {
			Position prev = fInternalMap.put(annotation, position);
			if (prev == position) {
				return prev;
			}
			if (prev != null) {
				fPositions.remove(prev);
			}
			fPositions.put(position, annotation);
			return prev;
		}
	}

	@Override
	public Position get(Object annotation) {
		synchronized (fLockObject) {
			return fInternalMap.get(annotation);
		}
	}

	public Annotation getAnnotation(Position p) {
		synchronized (fLockObject) {
			return fPositions.get(p);
		}
	}

	@Override
	public void clear() {
		synchronized (fLockObject) {
			fInternalMap.clear();
			fPositions.clear();
		}
	}

	@Override
	public Position remove(Object annotation) {
		synchronized (fLockObject) {
			Position prev = fInternalMap.remove(annotation);
			if (prev != null) {
				fPositions.remove(prev);
			}
			return prev;
		}
	}

	@Override
	public int size() {
		synchronized (fLockObject) {
			return fInternalMap.size();
		}
	}

	@VisibleForTesting
	int countPositions() {
		synchronized (fLockObject) {
			return fPositions.size();
		}
	}

	@Override
	public boolean isEmpty() {
		synchronized (fLockObject) {
			return fInternalMap.isEmpty();
		}
	}

	/**
	 * @deprecated Should not be called and will always throw an {@link UnsupportedOperationException}.
	 */
	@Deprecated(forRemoval = false)
	@Override
	public boolean containsValue(Object value) {
		throw new UnsupportedOperationException();
	}

	/**
	 * @deprecated Should not be called and will always throw an {@link UnsupportedOperationException}.
	 */
	@Deprecated(forRemoval = false)
	@Override
	public void putAll(Map<? extends Annotation, ? extends Position> map) {
		throw new UnsupportedOperationException();
	}

	/**
	 * @deprecated Should not be called and will always throw an {@link UnsupportedOperationException}.
	 */
	@Deprecated(forRemoval = false)
	@Override
	public Set<Entry<Annotation, Position>> entrySet() {
		throw new UnsupportedOperationException();
	}

	/**
	 * @deprecated Should not be called and will always throw an {@link UnsupportedOperationException}.
	 */
	@Deprecated(forRemoval = false)
	@Override
	public Set<Annotation> keySet() {
		throw new UnsupportedOperationException();
	}

	/**
	 * @deprecated Should not be called and will always throw an {@link UnsupportedOperationException}.
	 */
	@Deprecated(forRemoval = false)
	@Override
	public Collection<Position> values() {
		throw new UnsupportedOperationException();
	}
}
```

Of course failing now since entrySet() is being called :p

Also changes to (all subtypes of) AnnotationModel had to be done. It's just a very sad data structure that they built there.

mickaelistria: > Should we get rid of IAnnotationMap.... in theory, yes.

I think that would be a welcome change in the end if we can stick to JVM types. One could introduce an `AnnotationModel.getAnnotationsAsMap()` or similar method returning another type, make sure `AnnotaitonModel.getAnnotationMap()` is not used any more in Platform/JDT/PDE, and then deprecate both the method and IAnnotationMap

> In another context I rewrote the better part of AnnotationModel and friends, but that's not strictly backwards compatible. Not sure really how to deal with it.

It's always very tricky. If you can find increments that are not breaking, you can try to get them in; so we see the improvements and it may motivate to start considering more serious other API proposal and progressively deprecate current one in favor of something better.

szarnekow: Products that run on various Eclipse versions would not benefit from these since the old IAnnotationModel interface must still be implemented correctly according to the usages of all supported Eclipse versions. Don't get me wrong here: The backwards compatibility is a great asset and should not be discarded lightheartedly. I'm only seeing the regression because my local patch was incomplete.

Issue #23, opened 2022-05-12
Hover on hit Breakpoint relayouts wrong on entering


tsmaeder:
When I hit a breakpoint, there are two markers in the annotation bar: the breakpoint and the current instruction pointer. When I hover over the breakpoint, I get a popup displaying some information. The `information` being displayed is

```
<html><body text="#000000" bgcolor="#ffffe1"><p>Multiple markers at this line<ul><li>Debug Current Instruction Pointer</li><li>Line breakpoint:Float [line: 458]</li></ul></body></html>
```

rendering as 
```
Multiple markers at this line
  * Debug Current ....
```
When you hover over the popup, the popup is being displaced with an window that has focus and can be resized. However, the first line is not displayed as before, instead it shows as 
```
Multiple markers at this 
line
...
```

This hides a good portion of the popup text.

The trouble seems to be this: if we take the size of the control, then reapply that as in DefaultInformationControl.setSizeConstraints(), the text is layouted differently.
My suspicion is that the `INNER_BORDER` is not taken into account when computing the necessary size for the original control. When there is only a single line in the popup, the replacement works without breaking the first line.

merks:
In my experience, the calculations that determine the width/size of some HTML text to display in a hover is horribly inaccurate and at best can be considered a heuristic.    :-(

tsmaeder:
Also in the `LineBreakingReader` class: 

```
	 * Reads the next line. The lengths of the line will not exceed the given maximum width.
	 *
	 * @return the next line
	 * @throws IOException if an I/O error occurs
	 */
	public String readLine() throws IOException {
		if (fLine == null) {
			String line= fReader.readLine();
			if (line == null)
				return null;

			int lineLen= fGC.textExtent(line).x;
			if (lineLen < fMaxWidth) {
				return line;
			}
```
if we're looking for "will not exceed the given maximum width", we should compare with `<=`, not `<':
```
if (lineLen < fMaxWidth) {
```

Fixing the comparison fixes the problem with the first line.

tsmaeder:
Using a text layout is blocked by https://github.com/eclipse-platform/eclipse.platform.swt/issues/184

PR #26, opened 2022-05-13
Bug 562709: Fix layout for focused quickfix popup

tsmaeder:
### What i does

This is a fix for [Bug 562709](https://bugs.eclipse.org/bugs/show_bug.cgi?id=562709) - Focused quick fix popup doesn't properly layout quick fix options. 
The problem arises from the fact that some IInformationControl implementations need a size constraint set in order to properly layout themselves when an input is set, but the size of the shell needed to show a given content may change based on input (because the toolbar in the status row of the popup in this instance gets filled based on the input).
The fix consist of first setting a size constraint, then setting the input, and then computing the size again with the input set.

### How to test
Invoke the various information controls that can be replaced (i.e. that can take focus and become resizable) and make sure the contents are properly displayed in the focused state.  Examples are:

* Editor annotation bar hover 
* Javadoc hover
* Quickfix hover
* Debug value hover

I have tested the fix on Fedora 35 and Windows 11. If someone could test on a Mac, that would be helpful.

Note that there is another bug regarding text layout, which has different causes and is described here: https://github.com/eclipse-platform/eclipse.platform.text/issues/23

### DIFF
--- a/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/InformationControlReplacer.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/InformationControlReplacer.java
@@ -131,37 +131,43 @@ public class InformationControlReplacer extends AbstractInformationControlManage
 	public void showInformationControl(Rectangle subjectArea, Object information) {
 		IInformationControl informationControl= getInformationControl();
 
-		Rectangle controlBounds= fContentBounds;
-		if (informationControl instanceof IInformationControlExtension3) {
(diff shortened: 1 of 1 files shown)


vogella:
Best PR description I have seen in a long time combined with a very short commit message. :-) Please move the PR description to the commit message. AFAIK the PR description uses per default the last commit message, so by putting this into the commit message, so also get a great PR description for free.

mickaelistria:
Thanks. It does indeed make sense to recompute the best size of the popup when focusing on it.

PR #30, opened 2022-05-20
Improve FileSearch performance (fixes #29)

### LINKED ISSUE #29: Filesearch: each file is read twice
Search could be improved: The file content should be cached:
image

jukzi:
* Read each file only once (if it fits in the buffers) => ~ 30% faster

* Immediate report search result while still searching => faster visible

* Share workload between threads to avoid that single job with huge file
is running much longer then the others => faster finished.

* Set current filename as job name => better progress report.

### DIFF
--- a/org.eclipse.search/META-INF/MANIFEST.MF
+++ b/org.eclipse.search/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.search; singleton:=true
-Bundle-Version: 3.14.100.qualifier
+Bundle-Version: 3.14.200.qualifier
 Bundle-Activator: org.eclipse.search.internal.ui.SearchPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/org.eclipse.search/search/org/eclipse/search/internal/core/text/FileCharSequenceProvider.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/core/text/FileCharSequenceProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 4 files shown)


mickaelistria:
This change seems to introduce some risk on a core feature and is IMO not safe enough to put it in RC1. OK to move it to 4.25 M1 ?

iloveeclipse:
> This change seems to introduce some risk on a core feature and is IMO not safe enough to put it in RC1. OK to move it to 4.25 M1 ?

This is neither a bug fix for an urgent issue nor something that was broken recently, so 4.24 is surely wrong anyway.

jukzi:
sure this can wait for M1

HannesWell:
After I upgraded to 4.25 M1 today I got the following reproducible error for the first time when searching the workspace.
I searched for the simple string `PublicKey` using the file name pattern `*.java`. All other options are disabled.

```
java.lang.NullPointerException
	at org.eclipse.search.internal.core.text.TextSearchVisitor$ReusableMatchAccess.getFileContentLength(TextSearchVisitor.java:125)
	at org.eclipse.search.internal.ui.text.FileSearchQuery$TextSearchResultCollector.getLineElement(FileSearchQuery.java:131)
	at org.eclipse.search.internal.ui.text.FileSearchQuery$TextSearchResultCollector.lambda$0(FileSearchQuery.java:97)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1908)
	at org.eclipse.search.internal.ui.text.FileSearchQuery$TextSearchResultCollector.acceptPatternMatch(FileSearchQuery.java:94)
	at org.eclipse.search.internal.core.text.TextSearchVisitor$TextSearchJob.processFile(TextSearchVisitor.java:234)
	at org.eclipse.search.internal.core.text.TextSearchVisitor$TextSearchJob.run(TextSearchVisitor.java:185)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
I suspect this is connected to the associated PR.
Nevertheless the search is very faster now 👍🏽

jukzi:
> After I upgraded to 4.25 M1 today I got the following reproducible error for the first time when searching the workspace.

Please attach "the workspace" or a minimal workspace where the error occurs. i can not reproduce in my workspace.

HannesWell:
> > After I upgraded to 4.25 M1 today I got the following reproducible error for the first time when searching the workspace.
> 
> Please attach "the workspace" or a minimal workspace where the error occurs. i can not reproduce in my workspace.

It was in my workspace that only contains the [Eclipse-Passage Project](https://github.com/eclipse-passage/passage) and was provisioned via their Oomph setup.

Unfortunately I cannot reproduce it anymore either.

But from looking at the code and with the Exception-Stacktrace, I think the only way this could happened is if in `TextSearchJob.processFile()` the `charsequence` String is null, which is the case if `getOpenDocument()` returns not null?

jukzi:
i guess
```
diff --git a/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java b/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
index 5764f60..ec55207 100644
--- a/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -205,13 +205,13 @@
 				}
 
 				List<TextSearchMatchAccess> occurences;
-				CharSequence charsequence = null;
+				CharSequence charsequence;
 
 				IDocument document= getOpenDocument(file, getDocumentsInEditors());
 				if (document != null) {
-					DocumentCharSequence documentCharSequence = new DocumentCharSequence(document);
+					charsequence = new DocumentCharSequence(document);
 					// assume all documents are non-binary
-					occurences = locateMatches(file, documentCharSequence, matcher, monitor);
+					occurences = locateMatches(file, charsequence, matcher, monitor);
 				} else {
 					try {
 						charsequence = fileCharSequenceProvider.newCharSequence(file);

```
would fix that, but i would like to have a situation to test it  - A file open in a text editor linked two times into the workspace? i tried with JDTs org.eclipse.jdt.internal.compiler.lookup.AnnotationHolder but could not reproduce.

iloveeclipse:
> A file open in a text editor linked two times into the workspace?

May be file from nested projects, or from linked virtual folder? There are few legal ways to have same resource appearing multiple times int the workspace.

jukzi:
can reproduce when i link a file external of workspace 2 times and have "both" file editors open.

PR #37, opened 2022-06-03
Use TextLayout to break hover lines. Fixes #23

### LINKED ISSUE #23: Hover on hit Breakpoint relayouts wrong on entering

tsmaeder:
# What it does

Fixes #23 by using a `TextLayout` instead of a `GC` to properly compute line wrapping in the default hover control and in the Javadoc view. 

#How to test
I've used the annotation bar hover (as mentioned in the issue) and the Javadoc view and Javadoc hover to test. In order to do that in the Java stuff (where it's only used as a fallback for the Browser widget), I changed the appropriate code to _not_ use the browser widget in the following places: 
```
JavaDocView.internalCreatePartControl
JavaDocHover.doCreateInformationControl();
```


tsmaeder:
Not sure who should look at this, but maybe @mickaelistria has an idea who to request for a review?

mickaelistria:
It's definitely a good thing to rely on TextLayout here, and the change is pretty internal. So let's merge it so we get time to play with it and decide of necessary fixes or revert before next release.

PR #40, opened 2022-06-06
552728: remove white artifacts

betamaxbandit:
When in dark theme, the Search dialog showed white artifacts when
switching between the search tabs.
Rather than creating each tab's content when the tab is clicked, all the
tab contents are created on dialog creation. This fixes the problem.

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchDialog.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchDialog.java
@@ -434,13 +434,11 @@ public class SearchDialog extends ExtendedDialogWindow implements ISearchPageCon
 			if (imageDesc != null)
 				item.setImage(imageDesc.createImage());
 
-			if (i == fCurrentIndex) {
-				Control pageControl= createPageControl(folder, descriptor);
-				pageControl.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
-				item.setControl(pageControl);
-				fCurrentPage= descriptor.getPage();
-				fDialogSettings.put(STORE_PREVIOUS_PAGE, descriptor.getId());
-			}
+			Control pageControl = createPageControl(folder, descriptor);
+			pageControl.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+			item.setControl(pageControl);
+			fCurrentPage = descriptor.getPage();
+			fDialogSettings.put(STORE_PREVIOUS_PAGE, descriptor.getId());
 		}
 
 		folder.addSelectionListener(new SelectionAdapter() {
@@ -545,11 +543,6 @@ public class SearchDialog extends ExtendedDialogWindow implements ISearchPageCon
 		CTabFolder folder = item.getParent();
 
 		SearchPageDescriptor descriptor= (SearchPageDescriptor) item.getData("descriptor"); //$NON-NLS-1$
-
-
-		if (item.getControl() == null) {
-			item.setControl(createPageControl(folder, descriptor));
-		}
 
 		Control oldControl= folder.getItem(fCurrentIndex).getControl();
 		Point oldSize= oldControl.getSize();


jonahgraham:
@betamaxbandit - as the master branch has just opened for next version development the version needs to be bumped:

```
17:22:22.358 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:3.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.search: Only qualifier changed for (org.eclipse.search/3.14.100.v20220606-1700). Expected to have bigger x.y.z than what is available in baseline (3.14.100.v20220120-1549) -> [Help 1]
```

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchDialog.java
```diff
@@ -434,13 +434,11 @@ protected Control createPageArea(Composite parent) {
...
 				item.setImage(imageDesc.createImage());
 
-			if (i == fCurrentIndex) {
-				Control pageControl= createPageControl(folder, descriptor);
-				pageControl.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
-				item.setControl(pageControl);
-				fCurrentPage= descriptor.getPage();
-				fDialogSettings.put(STORE_PREVIOUS_PAGE, descriptor.getId());
-			}
+			Control pageControl = createPageControl(folder, descriptor);
```
jonahgraham, line 437: I am concerned that calling createPageControl will activate every plug-in that contributes a search page rather than deferring such activation until the tab is actually requested.

betamaxbandit: Hi Jonah, thank you for taking a look at this.
I tried adding Control.requestLayout() on the tab folder, the control from createPageControl(...), the parent in createPageArea() and the active shell, but none of these had an effect.
I noticed the problem presented itself until all the tab content had been created, after which the white artifacts no longer appeared. Hence the rational for creating them all at the start.
However you make a good point about deferred activation. If this is unacceptable then I'll go back to the drawing board with this.

jonahgraham: Unless something has changed about the general requirements about activation I would guess that having a significant change of activation will be a problem.

Looking at this code a little bit it looks like the issue only happens in relation to the code if `showScope` is true - and in that block if you don't call `fScopeParts[index].setVisible(true);` then the problem goes away. The setVisible calls layout on the control, so it looks to me like a bug the first time layout is called. If so, that indicates there is some underlying bug that probably applies to numerous layouts. I guess the best case is to extract what that set of layouts are and see if it can be reproduced in a smaller set as that would fix this type of issue in multiple places in one go.

Note this is a place the SWT and layout spy can be problematic as it can display the result of current calculations, not the results that were actually used to display the gui.

jukzi:
bug link: https://bugs.eclipse.org/bugs/show_bug.cgi?id=552728
problem happens also on windows
image
image

Would be better if the theming issue would not require to touch the actual dialog

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchDialog.java
betamaxbandit: Hi,

As per @jonahgraham's suggestion, I've created a compact reproduction test case which demonstrates this problem.


The file Bug552728.zip is an Eclipse Plug-in Project. It contains a compact test reproduction case to demonstrate the bug:
Bug 552728 - [Dark Theme] Search shows white artifacts
To run test:
* Place the Bug552728 plugin in an Eclipse SDK workspace and launch a runtime workbench.
* In Preference > General > Appearance, change Theme to Dark and restart.
* In the main toolbar, select the menu Bug552728 and choose SearchDialogTest (or press Ctrl+6).
* A modal dialog appears containing 3 tab items, with item 0 selected.
* Click Item 1 tab:
Expected: The text "Content for Item" is shown in white, the remaining background area is drawn in the dark theme background colour.
Actual: The text "Content for Item" is shown in white with a small dark background and the remaining background area is drawn in white.


I've stepped through Composite.layout() and GridLayout.layout(Composite, boolean) and don't see anything obviously suspicious, although the code is insanely complex.

I would really appreciate it if someone familiar with SWT could take a look please.

Cheers John

[Bug552728.zip](https://github.com/eclipse-platform/eclipse.platform.text/files/10842484/Bug552728.zip)

laeubi:
This repository was merged with https://github.com/eclipse-platform/eclipse.platform.ui if your pull-request is still relevant please rebase the code and push it to the new repository.

PR #43, opened 2022-06-10
Replace filecontent dialog

vogella:


vogella:
@mickaelistria can you review / test? I did not find a way to trigger this dialog from a runtime IDE, creating files in the runtime IDE and changing it outside the IDE does not do the trick. Also changing files via Git does not seem to trigger the dialog in a runtime IDE:

mickaelistria:
I can reproduce it from a runtime IDE without issue: open a file in Eclipse IDE default text editor, right-click > Open With > System Editor (gedit for me), edit in Eclipse IDE so it becomes dirty, edit and save in gedit, come back to Eclipse IDE.

vogella:
Thanks @mickaelistria

PR #49, opened 2022-06-25
Improve async completion popup for completion on typing.

gayanper:
The async completion popup always shows the progress and the resulting
message of completions. The fix improves it so if async completion is
triggered automatically, then the popup is only shown if there are
completions to be shown.

### DIFF
--- a/org.eclipse.jface.text/META-INF/MANIFEST.MF
+++ b/org.eclipse.jface.text/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jface.text
-Bundle-Version: 3.20.200.qualifier
+Bundle-Version: 3.21.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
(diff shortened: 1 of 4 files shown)


vogella:
This would be really helpful to have less distractions in the user interaction. 

I currently can't test but from the description I understand that this would only be used for popups which are automatically triggered, if the user explicitly triggers it the popup would show in any case.

Definitely a big enhancement in the usabiltiy for me. Having an automatic empty popup showing that Eclipse is calculating something is annoying for me.

gayanper:
> I don't think it's an enhancement to not provide feedback on computation. If completion is long, it may give the impression to users that completion is not working, while something is computing, something that may be worth waiting for. Without a hint of progress, users miss a hint that there is a feature at play.

It's only done when auto activated rather than activated via shortcuts. IJ handles this by not running long running computations when auto activated, Even VSCode java don't have long running computation yet. Thats one reason now i have added the activation mode into JDT completion computer's context to improve this behavior.

But if we think progress is vital in auto activation mode, we could show the popup after a predefine time if still the results are getting generated if user stops typing. That will avoid the distraction while typing and also provide the necessary feedback if user is waiting for it.

vogella:
Can you fix the build error?

gayanper:
> Can you fix the build error?

Fixed, took a while to figure out the correct version strategy.

vogella:
Thanks @gayanper

iloveeclipse:
@gayanper : could you please look at https://github.com/eclipse-platform/eclipse.platform.ui/pull/1003 which says it is a fix for the regression introduced in this PR?

Issue #33, opened 2022-05-31
Copy-Paste doesn't trigger Linked Editing `IExitPolicy.doExit(...)`


vrubezhny:
When using a linked editing to simultaneously modify the linked editing session is to be stopped when typing characters that are not allowed in the context, for example when a Space character is typed while editing a method name.
To stop linked editing we can use the  `IExitPolicy` interface: 
```
org.eclipse.jface.text.link.LinkedModeUI.IExitPolicy.doExit(LinkedModeModel, VerifyEvent, int, int)
```
this interface allow us to process the characters typed in by a user. And adopters can implement their own schemas to exit a linked editing session,

However a document cab be modified not only by typing characters, for example, users may use Copy-Paste operation to change a text, including during the linked editing session is on. A Copy-Paste operation doesn't generate any `VerifyEvent` events and, as result, when the text is that is to be pasted contains any "not allowed" characters, all those characters will be "copied" into the linked edited text regions.

Steps to reproduce:
- Open or create a java file that has a method defined and/or some occurrences of use of that method.
- Select some text containing the characters not allowed to be treated as method name characters (for example, containing spaces, braces, comas etc).
- Start `Refactor-Rename` on the selected method 
- Paste the copied text.
- FAIL: The method name occurrences will be changed by adding the copied text to every method name occurrence within the text, The syntax will be broken as the result.

The current Linked Editing API doesn't prevent this issue, neither adopters can prevent it while using the existing API

iloveeclipse:
This caused over 20 new fails in https://download.eclipse.org/eclipse/downloads/drops4/I20221006-1800 on all platforms. 

Please investigate. 

See for example https://download.eclipse.org/eclipse/downloads/drops4/I20221006-1800/testresults/html/org.eclipse.jdt.text.tests_ep426I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

iloveeclipse:
One example :
```
java.lang.NullPointerException
at org.eclipse.jface.text.link.LinkedModeUI.connect(LinkedModeUI.java:1013)
at org.eclipse.jface.text.link.LinkedModeUI.enter(LinkedModeUI.java:755)
at org.eclipse.jdt.internal.ui.text.template.contentassist.TemplateProposal.apply(TemplateProposal.java:299)
at org.eclipse.jdt.text.tests.contentassist.PostFixCompletionTest.applyProposal(PostFixCompletionTest.java:932)
[...]

PR #55, opened 2022-07-07
Copy-Paste doesn't trigger Linked Editing IExitPolicy.doExit(...) #33

vrubezhny:
In order to allow using the IExitPolicy interface to control exiting the Linked Editing Mode
on Paste operations this fix suggests extentong the IExitPolicy interface with a new method.
The method should take a `DocumentEvent` as an argument and is to be used from `DocumentListener`
to check if we should exit the linked mode.

Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>
Change-Id: Ib75726963f9dd5eb2d70c24993d011dc48e9ce81


org.eclipse.jface.text/src/org/eclipse/jface/text/link/LinkedModeUI.java
```diff
@@ -279,14 +279,40 @@ public interface IExitPolicy {
...
+	/**
+	 * An extension to exit policy can be registered by a caller to get custom exit behavior for
+	 * external changes like Copy-Paste operations.
+	 * <p>
+	 * Clients may implement this interface.
+	 * </p>
+	 *
+	 * @since 3.21
+	 */
+	public interface IExitPolicyExt extends IExitPolicy {
```
mickaelistria, line 291: Instead of an `Ext` interface, we can now add `default` methods on existing interfaces and this is backward compatible. Can you please rework it to use that?
Also, it would be great if this could come with an automated test.

iloveeclipse: > Adding a method into an exiting interface requires us to bump Major version for the plugin

Try to add "default" method.

vrubezhny: @iloveeclipse I did it as @mickaelistria suggested... You can spot it in the error message:
```
The default method org.eclipse.jface.text.link.LinkedModeUI.IExitPolicy.doExit(LinkedModeModel, DocumentEvent) in an interface has been added
```
It doesn't matter what kind of method is added to an existing interface ("default"  or not) - it's all treated as API break and, as such, requires to bump the major version.

akurtakov: Would you please provide alternative patch with default method ? If what you say happens this is a bug in PDE and we should add api filter.

vrubezhny:
A variant of the fix that adds a `default` method to the existing `IExitPolicy` interface instead of adding a new interface: https://github.com/eclipse-platform/eclipse.platform.text/pull/59

PR #58, opened 2022-07-16
Undo operation fails to correctly restore text

vrubezhny:
The Linked Editing is to be disabled on Undo operatioin, otherwise some extra characters
may be removed/inserted by Undo operation.

Fixes: https://github.com/eclipse/lsp4e/issues/178

Change-Id: I51a921caa482262b0a73a73dba709e554f428a80
Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

### DIFF
--- a/org.eclipse.jface.text/src/org/eclipse/jface/text/link/LinkedModeUI.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/text/link/LinkedModeUI.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


mickaelistria:
Before changing the default, have you investigate whether JDT makes this work already and if yes, how?

mickaelistria:
We'll need a unit test for this, that reproduces the issue and verifies that patch fixes it.

mickaelistria (changes requested):
I don't think undoing should exit the linked edition, but instead it should ignore linked edition for the edits that are part of the undo operation.
eg if I have `<a></a>` and linked edit ranges cover both `a`, I type `b` so it becomes `<ab></ab>` then undoing should restore `<a></a>` but leave the two `a`s connected as linked edit ranges.

mickaelistria:
Also, maybe using `DocumentUndoManagerRegistry.getDocumentUndoManager(fDocument).addDocumentUndoListener(...)` is also a good way to know whether the current change is part of an undo or not.
Some strategy we may consider, once we know how to best monitor coumpound changes/undo/redo/...:
* If change affects a position that is outside of the linked edit ranges, then don't disable linked edit range
* If the change affects only 1 position inside the linked edit ranges, then apply it and propagate to other linked edit ranges
* If the change affect all linked edit ranges and makes that the ranges are still "equals", then apply it as it, skipping the linked edit, and keep the linked edit ranges enabled for further edition
* If the change affects multiple position in the linked edit ranges and the linked edit ranges have diverging content as a result, exit the linked edit mode.

vrubezhny:
The fix is changed s making the Linked Editing to be disabled on Undo operation  always when a text that is to be modified is fully or partially outside of Linked Editing region.
Otherwise it makes sure that the current compound change is correctly closed before the start of Undo operation.

mickaelistria (approved):

PR #60, opened 2022-07-22
Fix NPE when searching in open file which is linked multiple times

jukzi:
see https://github.com/eclipse-platform/eclipse.platform.text/pull/30#issuecomment-1185512753


HannesWell (approved):
This change looks good and fixed the NPE for me in the reproducible case you mentioned.

Thank you!

HannesWell:
M2 Stabilization starts on Tuesday midnight.
So if you are fine with this as well and want it in M2, it can be merged until Monday 23:59.

jukzi:
thanks @HannesWell for reporting and testing it too.

HannesWell:
Thank you for fixing this :)

Issue #65, opened 2022-07-28
"Unhandled event loop exception" on "Cancel user task"


iloveeclipse:
- Tried to edit a file during the long running build.
- Blocking progress dialog shows me, that I have to wait.
- I don't wait so click on "Cancel user operation"
- "Unhandled event loop exception" error logged to the error log

This is coming from https://bugs.eclipse.org/bugs/show_bug.cgi?id=577511 / https://github.com/eclipse-platform/eclipse.platform.text/commit/0f409d6f3d6b590242268c9d97c3a0107737dfa7 / https://github.com/eclipse-platform/eclipse.platform.text/pull/4.

If we re-throw this exception in order to fix something, we should not log it as "Unhandled event loop exception" **error**. A warning, if any, makes sense.

I assume org.eclipse.ui.application.WorkbenchAdvisor.eventLoopException(Throwable) should ignore or log OperationCanceledException instances, but might be there is a better idea.

```
eclipse.buildId=4.25.0.202207280832
java.version=17.0.3
java.vendor=Eclipse Adoptium
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -data file:/tmp/wsp_rnd_smoke/ -os linux -ws gtk -arch x86_64

org.eclipse.ui
Error
Thu Jul 28 11:41:09 CEST 2022
Unhandled event loop exception

org.eclipse.core.runtime.OperationCanceledException
	at org.eclipse.ui.texteditor.AbstractTextEditor.validateState(AbstractTextEditor.java:4906)
	at org.eclipse.ui.texteditor.AbstractTextEditor.lambda$4(AbstractTextEditor.java:4947)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74)
	at org.eclipse.ui.texteditor.AbstractTextEditor.validateEditorInputState(AbstractTextEditor.java:4947)
	at org.eclipse.ui.texteditor.StatusTextEditor.validateEditorInputState(StatusTextEditor.java:159)
	at org.eclipse.ui.texteditor.AbstractDecoratedTextEditor.validateEditorInputState(AbstractDecoratedTextEditor.java:1055)
	at org.eclipse.ui.texteditor.AbstractTextEditor$ElementStateListener$Validator.verifyText(AbstractTextEditor.java:392)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:277)
[...]

PR #72, opened 2022-08-09
Revert "bug 577511 - do not ignore cancel (#4)" (#64)

iloveeclipse:
This reverts commit 0f409d6f3d6b590242268c9d97c3a0107737dfa7.

Reason for revert: the original issue still there, but additionally we
have unhandled event loop exception.

Fixes https://github.com/eclipse-platform/eclipse.platform.text/issues/65


jukzi:
please hold on.

iloveeclipse:
I assume you've tested on Windows, and I assume breaking event loop there leads to different side effects as on Linux. 
On Linux I can easily reproduce original issue with/without patch.

iloveeclipse:
Superceeded by https://github.com/eclipse-platform/eclipse.platform.text/pull/73

Issue #79, opened 2022-08-24
StringIndexOutOfBoundsException during File Search


jukzi:
![image](https://user-images.githubusercontent.com/51790620/186518481-0c2ddf37-dbc6-4551-818c-554f1aa646b7.png)

```
!MESSAGE An internal error occurred during: "Searching for pattern 'Hello'...".
!STACK 0
java.lang.StringIndexOutOfBoundsException: String index out of range: 11
	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:47)
	at java.base/java.lang.String.charAt(String.java:693)
	at java.base/java.lang.Character.codePointAt(Character.java:8264)
	at java.base/java.util.regex.Pattern$CharPropertyGreedy.match(Pattern.java:4296)
	at java.base/java.util.regex.Matcher.match(Matcher.java:1756)
	at java.base/java.util.regex.Matcher.matches(Matcher.java:713)
	at org.eclipse.search.ui.text.FileTextSearchScope.matchesFileName(FileTextSearchScope.java:226)
	at org.eclipse.search.ui.text.FileTextSearchScope.contains(FileTextSearchScope.java:220)
	at org.eclipse.search.internal.core.text.FilesOfScopeCalculator.visit(FilesOfScopeCalculator.java:41)
	at org.eclipse.core.internal.resources.Resource.lambda$0(Resource.java:85)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:85)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:90)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:90)
	at org.eclipse.core.internal.watson.ElementTreeIterator.iterate(ElementTreeIterator.java:127)
	at org.eclipse.core.internal.resources.Resource.accept(Resource.java:94)
	at org.eclipse.core.internal.resources.Resource.accept(Resource.java:55)
	at org.eclipse.search.internal.core.text.FilesOfScopeCalculator.process(FilesOfScopeCalculator.java:56)
	at org.eclipse.search.core.text.TextSearchScope.evaluateFilesInScope(TextSearchScope.java:86)
	at org.eclipse.search.internal.core.text.TextSearchVisitor.search(TextSearchVisitor.java:430)
[...]

PR #82, opened 2022-08-26
fix StringIndexOutOfBoundsException during File Search #79

jukzi:
If same search is started in parallel.
java.util.regex.Matcher is not thread-safe.


jukzi:
`[API ERROR] File MANIFEST.MF at line 5: The service version is increased unnecessarily since either the major or minor or service version is already increased (location: /home/jenkins/agent/workspace/eclipse.platform.text_PR-82/org.eclipse.search/META-INF/MANIFEST.MF)`

PR #83, opened 2022-08-29
File search: Reliable and fast cancel

jukzi:
File search did not cancel it's worker jobs if the user canceled before
the monitorUpdateJob was running (scheduling it is no guarantee that is
already running).
* monitor job is now incorporated into the search thread.
* fixed expected number of files to scan
* do not exit search thread before all workers finished.
* cancel search if any of the workers canceled
* forward cancel immediately (do not wait 100ms)


jukzi:
split from https://github.com/eclipse-platform/eclipse.platform.text/issues/78#issuecomment-1228417806

jukzi:
```
org.eclipse.core.filebuffers.tests.FileBuffersForFilesInLinkedFolders.test16_3
expected:<1> but was:<0>
```

Issue #87, opened 2022-09-15
Provide .editorconfig support


angelozerr:
Today `.editorconfig` is supported directly in vscode, IJ, etc

It should be really nice that Eclipse IDE provides this support too. It exists a discussion at https://bugs.eclipse.org/bugs/show_bug.cgi?id=457046 

And I implemented a long time ago https://github.com/angelozerr/ec4e which works with genereic editor but cannot apply all settings.

laeubi:
@angelozerr do you think you can prepare a PR for this to contribute the code to eclipse? I'll then open a CQ for this,

mickaelistria:
The edition part seems to require TM4E. Do we want Platform to start depending on TM4E ? I personally wouldn't mind, but I think it should be discussed a bit more broadly.
If we don't want it, then things like langauge configuration and highlighting would have to be removed for the moment.

jukzi:
some questions: 
a) What happens when the .editorconfig contradicts the workspace settings?
b) What happens when the .editorconfig in the project folder contradicts the project settings?
c) When the projects setting are edited via UI will the .editorconfig reflect the change?
d) does "searching" for the .editorconfig use file system or eclipse workspace api?

angelozerr:
At first I would like to say that ec4e is based on https://github.com/ec4j/ec4j an editorconfig parser that I have implemented which can provide the required information that we need for IDE (AST offsets of section, fault tolerant parser,etc) compared to https://github.com/editorconfig/editorconfig-core-java which don't give the capability to give some offsets.

The official site https://editorconfig.org/ references today the ec4j.

The ec4e should work well with Generic Editor but not correctly with old editor. Here a demo where you can wizard in action, codeming, validation of .editorconfig and the most important apply of property:

![EditorConfigDemo](https://user-images.githubusercontent.com/1932211/190447190-a8183175-77bc-414a-ab98-c78c60ddebd6.gif)

> The edition part seems to require TM4E. Do we want Platform to start depending on TM4E ? I personally wouldn't mind, but I think it should be discussed a bit more broadly.

Indeed, perhaps TM4E should take care of that?

> a) What happens when the .editorconfig contradicts the workspace settings?

It is the .editorconfig which should have the priority. But any improvement are possible.

> b) What happens when the .editorconfig in the project folder contradicts the project settings?

Again, it is the .editorconfig which should have the priority. But any improvement are possible.

[...]

PR #88, opened 2022-09-16
Get IPreferenceStore from AbstractTextEditor with adapt mechanism

angelozerr:
Get IPreferenceStore from AbstractTextEditor with adapt mechanism


angelozerr:
This simple PR provides the capabaility to get IPreferenceStore from an AbstractTextEditor with adapt mechanism:

```java
ITextEditor = textEditor = ...
IPreferenceStore store = textEditor.getAdapter(IPreferenceStore.class);
```

It avoids using Java Reflection to access it:

```java
Field field = AbstractTextEditor.class.getDeclaredField("fPreferenceStore");
field.setAccessible(true);
IPreferenceStore store = (IPreferenceStore) field.get(textEditor);
```

I have 2 usecases where I will need to get IPreferenceStore from a ITextEditor:

 * .`editorconfig support` : the apply of `indent_style = tab` for a given editor must update the preference store of the editor:

```java
ITextEditor = textEditor = ...
IPreferenceStore store = textEditor.getAdapter(IPreferenceStore.class);
boolean spacesForTabs = (indent_style != tab)
if (oldSpacesForTabs != spacesForTabs) {
	store.firePropertyChangeEvent(EDITOR_SPACES_FOR_TABS, oldSpacesForTabs, spacesForTabs);
}
```
See:

 * https://github.com/eclipse-platform/eclipse.platform.text/issues/87

 * manage custom preferences for spaces according to a language for LSP4E (ex : having a xml spaces preferences for XML)

See:

* https://github.com/eclipse/wildwebdeveloper/issues/139#issuecomment-1075378084
 * https://github.com/eclipse/lsp4e/issues/245

laeubi (approved):
That's a good approach here!

laeubi:
@angelozerr the build failure is not related to your change, I'll take a look into it!

mickaelistria (changes requested):
I still have concerns with the idea of reusing the editor preference store while most of the possible operations may not require an editor but could instead start from an IDocument (think about JDT-LS).
We're discussing possible alternatives (or lack of) with @angelozerr and possible increments towards the bigger goal. I'd rather wait for some former conclusions to be reached before merging this.

angelozerr:
@mickaelistria here a concrete usecase for this PR. LSP4e formatter needs to get the preference store of the generic editor to get tabWidth / insert spaces for a given language

https://github.com/eclipse/lsp4e/pull/271/files#diff-32bb8d5387f288529fd4b881ee69dbdd3282c84cc9c452605ab770d968732478R74

angelozerr:
@mickaelistria here an new concrete sample with TM4e language configuration to get insert spaces / tab width when on enter rule (when user type Enter,it indent correctly and should get the tab with / insert spaces from the editor preference store

See  * and access to the editor preference store https://github.com/eclipse/tm4e/pull/461/files#diff-2961979865d7499073f384bc8a71472fb1d5298dd54d3c4713c8f6ed94405801R295

angelozerr:
Is there any chance to merge this PR? I tried to show you several usecases when IPreferenceStore must be get from an editor, and today we need to use Java reflection to access it, which is very bad.

laeubi:
As it is about editor-config I think it is fair to adapt an editor but have no idea about the IDocument approach @mickaelistria maybe you can outline a bit more how this is supposed to work?

I also checked the code-base a bit and I don't see that only *one* approach will work, for a full solution one would most probably need different approaches, e.g IDocument, Editors and even IFile as one can format source files also without any editor (and thus possibly any document).

@angelozerr can you rebase the PR and make sure the build works?

laeubi (approved):
I think the change is not harmful and do not add any special maintenance burden.
Even if there are better ways to archive this it could be a good first step.

angelozerr:
> As it is about editor-config I think it is fair to adapt an editor but have no idea about the IDocument approach @mickaelistria maybe you can outline a bit more how this is supposed to work?

It is not only for .editorconfig support. The main idea is that an ITextEditor is created with preference store and some external component need to access to those preferences. 

Take a sample without .editorconfig:. In WWD we have more and more users request who want to have the capability to customize spaces for XML (and don't use the global EditorUI preferences, like WTP XML editor). An external component created outside the WWD (like LSP4E, TM4E)  needs to access to this preference to support format, auto editing strategies, etc, and they don't know WWD and which preferences store which is used to manages spaces.

This PR gives the capability to get a preference store registered in the editor without using Java Reflection because AbstractTextEditor#getPreferenceStore is protected.

I have several usescase to use this PR:

 * LSP4E : https://github.com/eclipse/lsp4e/pull/271/files#diff-32bb8d5387f288529fd4b881ee69dbdd3282c84cc9c452605ab770d968732478R74
 * TM4E : https://github.com/eclipse/tm4e/issues/460
 * .editorconfig support

angelozerr:
> @angelozerr can you rebase the PR and make sure the build works?

rebased

laeubi:
All checks are green, please @mickaelistria either adjust your review or give guidance how we can proceed here. My vote would be to merge this and if we later find better alternative it even is not harmful. Maybe also @vogella likes to give another review on this.

vogella:
LGTM

mickaelistria:
What about making `getPreferenceStore()` public?

angelozerr:
> What about making getPreferenceStore() public?

 - it will change the API
 - it will require to cast the ITextEditor into AbstractTextEditor to get the preference store
 - it is not consistent with ITextViewer (you can get ITextViewer with getAdapter and getSourceViewer is protected)

mickaelistria:
While I don't think those arguments are really strong; and that not publishing an API and instead relying on unspecified behavior is ugly; I looked at the code, and it really seems like there is a real effort to not expose such implementation details. 
So let's stick to that (anti-)pattern for the moment.

PR #89, opened 2022-09-17
Provide .editorconfig support (POC)

### LINKED ISSUE #87: Provide .editorconfig support

angelozerr:
Provide .editorconfig support (POC)

Fixes #87


angelozerr:
Here a POC with .editorconfig without TM4E (you have not syntax coloratuon and language configuration). 

It is a POC so I know that there a are a lot of issues (bad comments, I use ec4j JAR, etc), but I would like to start some discussion.

With editorconfig context we need to:

 * get the IPreferenceStore from an ITextEditor (outside the editor himself)
 * add an EditorConfigPreferenceStore on the first chained preference store of an editor.

I did that with Java reflection at https://github.com/angelozerr/ec4e/blob/80db2809069f9f7dc74cf60d94148c4af7aec3dd/org.eclipse.ec4e/src/main/java/org/eclipse/ec4e/internal/PreferenceStoreHelper.java#L33

I try to avoid using Java Reflection 

 * get the IPreferenceStore from an ITextEditor with adapter https://github.com/eclipse-platform/eclipse.platform.text/pull/89/files#diff-1a74925ffca65cb639987b023a79a1ada855cfd0b03e7de4196718f78d6609c1R6178 I created a simple PR with this idea at https://github.com/eclipse-platform/eclipse.platform.text/pull/88
 
 * add an EditorConfigPreferenceStore on the first chained preference store of an editor. To do that I created a new extension point for generic editor preferenceStoreFactory used by .editorconfig her 
 * https://github.com/eclipse-platform/eclipse.platform.text/pull/89/files#diff-3cf3f500075191dddbd8c67e1983f5994b44156356ea231e0263d9337440804dR176

This extension point requires to expose a new API IPreferenceStorefactory in genereiceditor, @mickaelistria not sure you will like that.

In another words I would like to fix this 2 problems with IPreferenceStore in order to I can continue to work on this POC with clean mean.

Many thanks for your feedback!

laeubi:
@angelozerr can you extract the generic editor enhancement into a separate PR?

Issue #34, opened 2022-06-02
No code mining updated when using Ctrl+End (was: CodeMiningTest are failing on Windows and Mac)


ktatavarthi:
This is a duplicate of [Bug 571317](https://bugs.eclipse.org/bugs/show_bug.cgi?id=571317) - No code mining updated when using Ctrl+End (was: CodeMiningTest are failing on Windows and Mac).

I will close the bug in bugzilla and this bug will be tracked here.

The below JFace text tests in org.eclipse.jface.text.tests.codemining.CodeMiningTest  fail in the nightly build on windows

1) testCodeMiningDoesntAlterFocus
2) testCodeMiningCtrlEnd
3) testCodeMiningCtrlHome
4) testCodeMiningFirstLine

ktatavarthi:
The below tests failed in the IBuild  [I20220601-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220601-1800/testResults.php) on [Windows](https://download.eclipse.org/eclipse/downloads/drops4/I20220601-1800/testresults/html/org.eclipse.jface.text.tests_ep424I-unit-win32-java11_win32.win32.x86_64_11.html)
- testCodeMiningDoesntAlterFocus
- testCodeMiningCtrlEnd
- testCodeMiningCtrlHome
- testCodeMiningFirstLine

ktatavarthi:
The below tests failed in the IBuild [I20220817-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220817-1800/) on [Windows](https://download.eclipse.org/eclipse/downloads/drops4/I20220817-1800/testresults/html/org.eclipse.jface.text.tests_ep425I-unit-win32-java11_win32.win32.x86_64_11.html)
- testCodeMiningDoesntAlterFocus
- testCodeMiningCtrlEnd
- testCodeMiningCtrlHome
- testCodeMiningFirstLine

ktatavarthi:
The below tests failed in the IBuild [I20220824-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20220824-1800/) on [Windows](https://download.eclipse.org/eclipse/downloads/drops4/I20220824-1800/testresults/html/org.eclipse.jface.text.tests_ep425I-unit-win32-java11_win32.win32.x86_64_11.html)
- testCodeMiningDoesntAlterFocus
- testCodeMiningCtrlEnd
- testCodeMiningCtrlHome
[...]

PR #96, opened 2022-09-30
fix CodeMiningProjectionViewerTest.testCodeMiningDoesntAlterFocus() #34

jukzi:
fResolvedMinings is filled only after draw.


mickaelistria:
This change seems correct. Please bump version by the way and we can merge ASAP to see how it behaves in real life.

PR #99, opened 2022-10-07
Fixes for 24 new test fails in jdt.text.tests

iloveeclipse:
IDocumentUndoManager is not always registered for a document. DocumentUndoManagerRegistry.getDocumentUndoManager() can return null, don't unconditionally dereference return value.

See https://github.com/eclipse-platform/eclipse.platform.text/issues/33


akurtakov:
Thanks Andrey!

PR #106, opened 2022-10-23
Leverage Tycho-pomless and make all pom.xml but the root obsolete

HannesWell:


HannesWell:
This requires 
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/645

akurtakov:
Many test failures so this one needs further work.

HannesWell:
> Many test failures so this one needs further work.

Erroneously I added the pom.model.property entries to the `plugin.properties` instead of the `build.properties`. Additionally the service version segments of the affected test plug-ins where not bumped.
All that is fixed now and I checked that the number of modules and tests remained the same. So this is ready to go.

Issue #85, opened 2022-09-13
Register IReconcilingStrategy with GenericEditor extension point.


angelozerr:
Todau GenericEditor provides the capability to register IReconciler with `org.eclipse.ui.genericeditor.reconcilers`

```xml
<extension
         point="org.eclipse.ui.genericeditor.reconcilers">
      <reconciler
            class="org.eclipse.lsp4e.operations.folding.LSPFoldingReconciler"
            contentType="org.eclipse.core.runtime.text">.
```

 `org.eclipse.lsp4e.operations.folding.LSPFoldingReconciler` extends MonoReconciler to benefit with call of reconcile methods with delay to update foldings. The main problem with this strategy that LSPFoldingReconciler create a brackgournd Thread which compute region and dirty region. Another recocniler registered will do that too.

In other words if there are 5 recocniler registered for a given content type, when one file is opened, it will create 5 background Thread and will will do the same process to compute the region and dirty region.

To fix this problem, GenericEditor should manag einternally a Reconciler to provide the capability to register IReconcilingStrategy with extension point:

```xml
<extension
         point="org.eclipse.ui.genericeditor.reconcilers">
      <reconcilingStrategy
            class="org.eclipse.lsp4e.operations.folding.LSPFoldingReconcilingStrategy "
            contentType="org.eclipse.core.runtime.text">.
```

The LSPFoldingReconcilingStrategy  requires to get:

 * the IDocument (IReconcilingStrategy#setDocument will give the capability to get the IDocument)
 * the viewer.

To intercept the viewer,  we could provide a new interface

```java
public interface ITextViewerLifecycle {

   void install(ITextViewer viewer)

   void uninstall();
}

```

In other words, the LSPFoldingReconcilingStrategy could looks liks this: 

```java
import org.eclipse.jface.text.IDocument;
import org.eclipse.jface.text.IRegion;
import org.eclipse.jface.text.ITextViewer;
import org.eclipse.jface.text.reconciler.DirtyRegion;
import org.eclipse.jface.text.reconciler.IReconciler;
import org.eclipse.jface.text.reconciler.IReconcilingStrategy;
public class LSPFoldingReconcilingStrategy  implements ITextViewerLifecycle, IReconcilingStrategy {

	private IDocument document;
	private ITextViewer viewer;

	@Override
	public void reconcile(IRegion arg0) {
		// use viewer and document to do some process with delay
	}

	@Override
[...]

PR #107, opened 2022-10-27
Introduce ITextViewerLifecycle API

angelozerr:
Introduce ITextViewerLifecycle API

As there are several API which defines install/uninstall methods for ITextViewer, I think it can be nice to promote that in a given API.

But my another motivation is to fix this issue at https://github.com/eclipse-platform/eclipse.platform.text/issues/85


laeubi:
@angelozerr you need to fix the API errors:
```
2 API ERRORS
[API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 3.21.100, since new APIs have been added since version 3.21.0 - The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.reconciler.Reconciler
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.text.contentassist.IContentAssistant
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.text.information.IInformationPresenter
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.contentassist.ContentAssistant
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.contentassist.ISubjectControlContentAssistant
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.reconciler.AbstractReconciler
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.information.InformationPresenter
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.codemining.CodeMiningReconciler
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.presentation.PresentationReconciler
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.text.presentation.IPresentationReconciler
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.text.hyperlink.IHyperlinkPresenter
- The type org.eclipse.jface.text.ITextViewerLifecycle has been added to org.eclipse.jface.text_3.21.100
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.text.reconciler.IReconciler
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.contentassist.SubjectControlContentAssistant
- The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for class org.eclipse.jface.text.reconciler.MonoReconciler (location: /home/jenkins/agent/workspace/eclipse.platform.text_PR-107/org.eclipse.jface.text/META-INF/MANIFEST.MF)
[API ERROR] File ITextViewerLifecycle.java at line 20: Missing @since tag on org.eclipse.jface.text.ITextViewerLifecycle (location: /home/jenkins/agent/workspace/eclipse.platform.text_PR-107/org.eclipse.jface.text/src/org/eclipse/jface/text/ITextViewerLifecycle.java)
```

angelozerr:
Thanks @laeubi for your feedback.

It should be fixed now.

laeubi (approved):
This looks good, but wait for @mickaelistria as he is our API-Guardian here :-)

angelozerr:
I update the since to `3.21.0` instead of `3.21`, is it correct?

> This looks good, but wait for @mickaelistria as he is our API-Guardian here :-)

Yes sure! As .editorconfig support is a big PR, I try to create little PR to speed the merge.

angelozerr:
@laeubi the build fails again, I'm a little lost, could you help me to fix it please. Thanks!

laeubi:
@angelozerr in general I search the (full) Console log for "[API ERROR] "

https://ci.eclipse.org/platform/job/eclipse.platform.text/job/PR-107/3/consoleFull

If you use the [Eclipse SDK](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#create-an-eclipse-development-environment) you should also see API error in the IDE, if not @merks might can help / advice why those not sho up.

org.eclipse.jface.text/src/org/eclipse/jface/text/ITextViewerLifecycle.java
```diff
@@ -0,0 +1,35 @@
...
+ *******************************************************************************/
+package org.eclipse.jface.text;
+
+/**
+ * {@link ITextViewer} lifecycle API to track install / uninstall of a given {@link ITextViewer}.
+ *
+ * @since 3.21.0
+ *
+ */
+public interface ITextViewerLifecycle {
```
szarnekow, line 22: After having read the JavaDoc I wonder where this API is being used. It doesn’t look like any check for the newly introduced interface happens anywhere. Naively I might assume that a random new service instance returned from e.g. a ISourceViewerConfig would be checked and the install would be called. But in fact this is not happening. Please add to the docs what an implementor can expect from adding this supertype and which services are allowed to be ITextViewerLifecycles

angelozerr: Ok I understand what you mean. My main motivation is to use ITextViewerLifecycle in generic editor https://github.com/eclipse-platform/eclipse.platform.text/issues/85 to track install / uninstall viewer inside generic editor contribution (auto editing stratetgy,etc). 

When I wanted to provide this API inside generic editor, I noticed that several API used install / unisntall methods with the same signature, so instead of providing this API in generic editor, I provide this API in JFace text directly.

In other words, this API doesn't provide a given service but it exists just to standardize the method name when you need to track install / uninstall.

angelozerr:
@szarnekow I updated my PR to use ITextViewerLifecycle with generic editor.

I created a PR on LSP4e side which consumes this API https://github.com/eclipse/lsp4e/pull/272

org.eclipse.jface.text/src/org/eclipse/jface/text/ITextViewerLifecycle.java
angelozerr: I think more and more that ItextViewerLyfecicle https://github.com/eclipse-platform/eclipse.platform.text/pull/107 should be hosted in generic editor

@mickaelistria @szarnekow what do you think about that?

szarnekow: To me this would be reasonable. Maybe also rename to IGenericTextViewerLifecycle or something more suited?

angelozerr: @szarnekow I move ITextViewerLifecycle to genereic editor and I keep the name ITextViewerLifecycle

mickaelistria:
I actually likes the first version of the patch better, particularly the part that factors the ITextLifecycle in existing reconcilers or others as it already emphasizes a lot of benefit in this factorization.
But I agree with @szarnekow that this new API should be made most obviously useful in general if we want it part of the common text API. Isn't there some other API that could optionally implement the provided interface, for example IQuickAssistProcessor or something like that? If we can show in the existing layer that this API can provide value to existing code, then it can make it a good candidate to get directly into jface.text. I suspect that about every possible strategy in `SourceViewer.configure()` can benefit from this API.

mickaelistria:
Just to be clear, what is the purpose of this API? Is it to allow more of existing contributions that support and install/uninstall pattern just like Reconcilier or ContentAssistProcessor already do, or to provide something new? If it's the former, I don't think you need more logic than adding `if (x instanceof ITextViewerLifecycle lifecycle) lifecycle.install(this)` here and there.

angelozerr:
> Is it to allow more of existing contributions that support and install/uninstall pattern just like Reconcilier or ContentAssistProcessor already do, or to provide something new? 

The both, but for LSP4e usecase, I would say "provide something new"

angelozerr:
> Just to be clear, what is the purpose of this API? 

My main goal is to track install / uninstall of ITextViewer for any contribution of genereic editor. I mean for IReconciler (which have already install / uninstall), but too for IAutoEditingStrategy (which have not install / uninstall methods). 

When you write a custom TextEditor (without generic editor), you need sometimes ITextEditor or track install / unisntall of ITextViewer. It is easy to do that since you can override SourceViewer or another component.

With generic editor you cannot do that, it is the main goal of ITextViewerLifecycle and the another API ITextEditorAware https://github.com/eclipse-platform/eclipse.platform.text/pull/108

Now the question is where host those 2 APIs?

 * in generic editor, it means that generic editor requires to export those API (today generic editor uses existing API like IReconciler)
 * in an another plugin like JFace Text, I tried to use it with existing component but according [this comment](https://github.com/eclipse-platform/eclipse.platform.text/pull/107#discussion_r1007844308) it seems that it is a bad idea. For ITextViewerLifecycle the main idea is to store all contribution which implements ITextViewerLifecycle  where contribution is created, and when source view install / unistall the TextViewer, we need to loop for ITextViewerLifecycle. I tried to do that for existing component in https://github.com/eclipse-platform/eclipse.platform.text/blob/d04589f174cac85f45fd4418dfa6053cbc16caa9/org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java#L492 but my fear is that I broke something, because there is some order of install and some updat eof boolean flag. I think it can be dangerous to do that.

In other words I think more and more that ITextViewerLifecycle and ITextEditorAware should be hosted in genereic editor plugin.

I'm wiating for your feedback to clean the both APIs.

mickaelistria:
> t according https://github.com/eclipse-platform/eclipse.platform.text/pull/107#discussion_r1007844308 it seems that it is a bad idea.

I don't think the comment says the idea is a bad one, but more that some documentation/usage of the new API is missing for it to feel complete. It's the same thing I tried to mention in my comment where I suggest we check for `instanceof ITextViewerLifecycle` on more objects of the SourceViewer.
The classes that can support this ITextViewerLifecycle to trigger the install/uninstall methods should receive a not in their documentation a bit like `*Extension` interfaces are documented.

> In other words I think more and more that ITextViewerLifecycle and ITextEditorAware should be hosted in genereic editor plugin.

The generic editor isn't meant to provide new Text APIs, JFace Text is the right place for APIs; Generic Editor just allows to contribute some extra behavior to the generic editor. The description of the use-case can make sense in other editors, use-cases than the generic editor. Usually the workaround is to pass directly a reference to the viewer when creating such a contributor, but it's actually not best for reusability, the suggested interface would also allow to improve existing cases that are independant from the generic editor.

angelozerr:
@mickaelistria is https://github.com/eclipse-platform/eclipse.platform.text/pull/107#issuecomment-1303964453 will foolow your idea?

mickaelistria:
> @mickaelistria is https://github.com/eclipse-platform/eclipse.platform.text/pull/107#issuecomment-1303964453 will follow your idea?

I don't think we need a new API method in SourceViewer for the moment, but just checking for `instanceof ITextViewerLifecycle` on the various objects that are used by this viewer (usually returned by the SourceViewerConfiguration). Already having it would validate the ITextViewerLifecycle addition.
Then, later in another change, we can consider adding a API method to add plain `ITextViewerLifecycle` that wouldn't be just decorations to existing elements, but I see it as a 2nd step when the ITextViewerLifecycle interface is already in place.

angelozerr:
@mickaelistria I push a new commit, please tell me if it is was your idea.

org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/AutoEditStrategyRegistry.java
```diff
@@ -62,16 +64,24 @@ public AutoEditStrategyRegistry() {
...
-			.collect(Collectors.toList());
+		return this.extensions.values().stream().filter(ext -> contentTypes.contains(ext.targetContentType))
+				.filter(ext -> ext.matches(sourceViewer, editor))
+				.sorted(new ContentTypeSpecializationComparator<IAutoEditStrategy>()) //
+				.map(ext -> {
+					IAutoEditStrategy delegateInstance = ext.createDelegate();
+					if (sourceViewer instanceof ISourceViewerExtension6
+							&& delegateInstance instanceof ITextViewerLifecycle) {
+						((ISourceViewerExtension6) sourceViewer)
+								.addTextViewerLifecycle((ITextViewerLifecycle) delegateInstance);
```
mickaelistria, line 80: The contributor (registry here) shouldn't have to explicitly call `viewer.addTextViewerLifecycle`, but instead the viewer should call `addTextViewerLifecycle` internally when installing the auto-edit strategies.

Let's look at TextHover for a more concrete proposal ⬇️

org.eclipse.jface.text/src/org/eclipse/jface/text/source/ISourceViewerExtension6.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.jface.text.source;
+
+import org.eclipse.jface.text.ITextViewerLifecycle;
+
+public interface ISourceViewerExtension6 {
```
mickaelistria, line 5: This shouldn't be necessary nowadays, we can instead consider using default methods in existing APIs.

org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
```diff
@@ -1320,4 +1316,23 @@ public void setCodeMiningAnnotationPainter(AnnotationPainter painter) {
 		ensureCodeMiningManagerInstalled();
 	}
 
+	@Override
+	public void addTextViewerLifecycle(ITextViewerLifecycle lifecycle) {
```
mickaelistria, line 1320: I don't think this have to be public at the moment. The viewer is responsible for calling install/uninstall on its "contributors", so this should have to be open.

org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ReconcilerRegistry.java
```diff
@@ -65,71 +70,85 @@ public ReconcilerRegistry() {
...
+	 * @param sourceViewer         the source viewer we're hooking completion to.
+	 * @param editor               the text editor
+	 * @param textViewerLifecycles the list of text viewer lifecycle to fill
+	 * @param contentTypes         the content types of the document we're editing.
 	 * @return the list of {@link IReconciler} contributed for at least one of the
 	 *         content types, sorted by most generic content type to most specific.
 	 */
 	public List<IReconciler> getReconcilers(ISourceViewer sourceViewer, ITextEditor editor,
-			Set<IContentType> contentTypes) {
+			List<IReconcilingStrategy> reconcilingStrategies, Set<IContentType> contentTypes) {
```
mickaelistria, line 81: I believe those changes are not so related to the plain addition of the interface and thus should be removed from the ocmmit.

org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/hover/TextHoverRegistry.java
```diff
@@ -89,32 +93,44 @@ public TextHoverRegistry(IPreferenceStore preferenceStore) {
...
+				// don't sort in the stream as the initial structure is already sorted by
+				// isAfter/isBefore
+				.map(ext -> {
+					ITextHover delegateInstance = ext.createDelegate();
+					if (sourceViewer instanceof ISourceViewerExtension6
+							&& delegateInstance instanceof ITextViewerLifecycle) {
+						((ISourceViewerExtension6) sourceViewer)
+								.addTextViewerLifecycle((ITextViewerLifecycle) delegateInstance);
+					}
+					return delegateInstance;
```
mickaelistria, line 112: The registry should not be responsible for calling the addTextViewerLifecycle here.
Instead, the CompositeTextHover should implement ITextViewerLifecycle and make the install/uninstall delegate to its children; and the ISourceViewer should check whether the provided ITextHover implements ITextViewerLifcycle and call install/uninstall on it when appropriate.

mickaelistria (changes requested):
Thanks.
This is already better as it shows how this is an opportunity for good factorization and how the pattern is often repeated and can be mimicked.
I've put some comments inline about the proposal. The main idea is that the "controbutors" are not responsible for calling install/uninstall. This must be invoked from the SourceViewer itself as it is the component that knows best when to call those according to its own state.

org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/hover/TextHoverRegistry.java
angelozerr: Thanks so much for your suggestion. I did that for IAutoEditingStrategy. I don't do that for ITextHover since ITextHover receives the ITextViewer.

The goal of ItextViewerLifecycle is to get the ItextViewer with a clean mean,as ITextHover receives the ITextViewer as parameter, I think it is not very helpful.

org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/ReconcilerRegistry.java
angelozerr: Removed

org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
angelozerr: Good catch! Now the method is private and ISourceViewerExtension6 has been removed.

org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/AutoEditStrategyRegistry.java
angelozerr: Fixed

org.eclipse.jface.text/src/org/eclipse/jface/text/source/ISourceViewerExtension6.java
angelozerr: Removed

org.eclipse.jface.text/src/org/eclipse/jface/text/ITextViewerLifecycle.java
angelozerr: @szarnekow I moved again ITextViewerLifecycle in JFace Text by foolowing great suggestion of @mickaelistria.

Now this API should be relevant. I updated the Javadoc to explain that. Hope you will like it.

angelozerr:
`[API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 3.21.100, since new APIs have been added since version 3.21.0 - The superinterfaces set has been expanded (org.eclipse.jface.text.ITextViewerLifecycle) for interface org.eclipse.jface.text.information.IInformationPresenter`

Could you help me please to fix that?

akurtakov:
It tells you to make https://github.com/eclipse-platform/eclipse.platform.text/blob/master/org.eclipse.jface.text/META-INF/MANIFEST.MF#L5 3.22.0.qualifier

org.eclipse.jface.text/src/org/eclipse/jface/text/ITextViewerLifecycle.java
```diff
@@ -0,0 +1,59 @@
...
+ * <li>{@link IAutoEditStrategy}</li>
+ * </ul>
+ *
+ * @since 3.22
+ *
+ */
+public interface ITextViewerLifecycle {
+
+	/**
+	 * Installs a text viewer.
```
szarnekow, line 49: Please clarify whether implementors should expect more than one install(..) call with / without uninstall invocations.

org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/MonoReconciler.java
```diff
@@ -98,4 +100,20 @@ protected void initialProcess() {
...
+	public void install(ITextViewer textViewer) {
+		super.install(textViewer);
+		if (fStrategy instanceof ITextViewerLifecycle) {
+			((ITextViewerLifecycle) fStrategy).install(textViewer);
+		}
+	}
+
+	@Override
+	public void uninstall() {
+		super.uninstall();
```
szarnekow, line 114: For the sake of symmetry, you might want to call the strategies uninstall before calling super.uninstall. Same is true for the some of the other impls.

org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
```diff
@@ -556,10 +559,22 @@ public void configure(SourceViewerConfiguration configuration) {
...
 		setCodeMiningProviders(configuration.getCodeMiningProviders(this));
-
+		installTextViewer();
 		activatePlugins();
 	}
 
+	private void doSetAutoEditStrategies(IAutoEditStrategy[] strategies, String contentType) {
+		super.setAutoEditStrategies(strategies, contentType);
+		if (strategies != null) {
+			for (int i= 0; i < strategies.length; i++) {
```
szarnekow, line 569: Preferably a `for (IAutoEditStrategy strategy: strategies)` instead of a counted for loop

szarnekow (approved):
Yes, I like it. This is the kind of comment and handling of the new API that I expected to happen.

mickaelistria:
@angelozerr I like it a lot as well in this current form. Please address @szarnekow 's comment and then we'll try to get this API addition merged ASAP (we're starting the RC phase, so we'll need some lobbying first but as it's backward compatible and the case is very well built with clear needs and various proposals were discussed to reach a consensus, I'm optimistic the development process and result are correct enough for a merge in RC).

org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
angelozerr: fixed

org.eclipse.jface.text/src/org/eclipse/jface/text/reconciler/MonoReconciler.java
angelozerr: fixed

org.eclipse.jface.text/src/org/eclipse/jface/text/ITextViewerLifecycle.java
angelozerr: I explain when install / uninstall is called. Hope it will be enough. If no please give me some suggestion.

angelozerr:
> It tells you to make https://github.com/eclipse-platform/eclipse.platform.text/blob/master/org.eclipse.jface.text/META-INF/MANIFEST.MF#L5 3.22.0.qualifier

I try, hope CI build will work.

angelozerr:
> @angelozerr I like it a lot as well in this current form.

Cool!

> Please address @szarnekow 's comment and then we'll try to get this API addition merged ASAP

I tried to fix all comments.

mickaelistria:
I don't get the API Tools issue here, it seems pretty wrong. Locally, I don't see this issue, either in the IDE nor in a Maven build with the right profiles triggered; the message gives me the impression an older version is used as baseline.

laeubi:
> I don't get the API Tools issue here, it seems pretty wrong. Locally, I don't see this issue, either in the IDE nor in a Maven build with the right profiles triggered; the message gives me the impression an older version is used as baseline.

@vik-chand can you help here?

laeubi:
Just from the code, maybe when an interface extends the new interface, instead of `@Override` them maybe remove them from the interface?

vik-chand:
>>vik-chand can you help here?

I fetched this pr on latest eclipse and I get
Description	Resource	Path	Location	Type
Invalid @since 3.21 tag on doExit(LinkedModeModel, DocumentEvent); expecting @since 3.22	LinkedModeUI.java	/org.eclipse.jface.text/src/org/eclipse/jface/text/link	line 295	@since tag problem

This is because the manifest changed the version and few weeks ago, there was an API added with since 3.21. So that should be changed to since 3.22 ( assuming we want this version)

So I skipped this error locally and I couldn't get further API tools error.

The expanded interface warning comes when an API interface has another interface added somewhere in its hierarchy and now available to its clients. I remember also making a change where in if the added interface breaks the client, then it will require a major version change.

If a change should report a error but doesn't in IDE, I think we need to fix it in IDE.
If a change reports an error in IDE but shouldn't, we need to fix it.
If something is not an error but this PR job reports it, then I guess @mickaelistria  may  know whats going on. It is possible that some baseline/profile/config is wrong but I'm not much of an expert there.  If not present in IDE, we can check in I build page ( which uses ant script) - https://download.eclipse.org/eclipse/downloads/drops4/I20221109-1850/apitools/analysis/html/index.html

mickaelistria:
@vik-chand thanks. I was looking at the commit date and not the merge date and didn't realize that the addition of `doExit` was *after* the release; what surprises me is that it was not noticed later, but now I believe it can be because an older version of parent was used by CI build and the "Rebase" part in "Rebase and merge" can have hidden the issue.
Anyway, @angelozerr , I think it should be `@since 3.22`. If you can add a commit that also fixes addition of `doExit` to `@since 3.22`, that would be welcome.

angelozerr:
@mickaelistria I'm little lost, I have `@since 3.22` in ItextViewerLifecycle. What I need to do ?

mickaelistria:
@angelozerr Can you please rebase on top of master and adapt the MANIFEST and `@since` tag to new 3.23 version? As soon as build is green, we can merge.

mickaelistria:
Thank you!

mickaelistria:
Please remember about adding a note to the N&N doc some time later (no need to rush)

org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
```diff
@@ -422,7 +428,7 @@ public SourceViewer(Composite parent, IVerticalRuler verticalRuler, IOverviewRul
 		fIsVerticalRulerVisible= (verticalRuler != null);
 		fOverviewRuler= overviewRuler;
 		fIsOverviewRulerVisible= (showAnnotationsOverview && overviewRuler != null);
-
+		this.lifecycles= new HashSet<>();
```
cdietrich, line 431: this seems to cause a memory leak in our Xtext Content assist test:
as the reconciler will now be uninstalled after the content assistant.
and thus additional JobInfos will be created for reconciler job

org.eclipse.jface.text/src/org/eclipse/jface/text/source/SourceViewer.java
```diff
@@ -712,18 +726,11 @@ public IAnnotationModel getVisualAnnotationModel() {
...
-			fPresentationReconciler.uninstall();
-			fPresentationReconciler= null;
-		}
-
-		if (fReconciler != null) {
-			fReconciler.uninstall();
-			fReconciler= null;
-		}
+		uninstallTextViewer();
+		fPresentationReconciler= null;
```
cdietrich, line 730: @angelozerr  @szarnekow my this nulling cause the problem we see in test?

PR #114, opened 2022-11-14
Bump o.e.jface.text version to 3.22

mickaelistria:
Previous change on LinkedModeUI.doExit actually added some API and should have triggered a version bump (but wasn't detected by API Tools).

PR #117, opened 2022-11-18
Filter out org.eclipse.jface.text version warning

mickaelistria:
An API was effectively added, LinkedModeUI$IExitPolicy.doExit, but some other filter are necessary and make API tool miss this

### DIFF
--- a/org.eclipse.jface.text/.settings/.api_filters
+++ b/org.eclipse.jface.text/.settings/.api_filters
@@ -1,5 +1,13 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <component id="org.eclipse.jface.text" version="2">
+    <resource path="META-INF/MANIFEST.MF">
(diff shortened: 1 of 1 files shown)


iloveeclipse:
For the house keeping: this is follow up on #114 version bump that was needed for #33

org.eclipse.jface.text/.settings/.api_filters
```diff
@@ -1,5 +1,13 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <component id="org.eclipse.jface.text" version="2">
+    <resource path="META-INF/MANIFEST.MF">
+        <filter comment="An API was effectively added, LinkedModeUI.doExit, but some other filter are necessary and make API tool miss this" id="926941240">
```
iloveeclipse, line 4: > but some other filter are necessary

Not sure I understood this sentence
It is important **if** that missing API error wan an API tools problem, report it in PDE.

iloveeclipse (approved):
Mickeal, for next PR's : it would be great for understanding where which change is coming from if you would add PR's or commit references to the related issues in the commit itself.

Issue #113, opened 2022-11-11
No way to configure ITextDoubleClickStrategy from extension point.


andrewL-avlq:
There does not appear to be an extension point for configuring the ExtensionBasedTextViewerConfiguration with a custom ITextDoubleClickStrategy (the strategy is hard coded to the default in the base class). Can a new extension point be added for this? Or is there a different way to customize double click behavior in the Generic Editor?

gerdleon:
@mickaelistria hi, we are looking at putting together a pull-request for this but wonder about the latest time this pull-request would still make it into the upcoming 4.26 release of Eclipse to be released on the 7th of December 2022 (see here: https://wiki.eclipse.org/Simultaneous_Release). When is the latest a pull-request can be approved for the upcoming release?

mickaelistria:
It's definitely too late for 4.26. We're already in RC phase and only fixes are supposed to be included until 4.26 is done and 4.27 is open for business.

mickaelistria:
But this will be a welcome change for 4.27!

gerdleon:
Thank you. What is the timeline for 4.27? Where can I read up on it?

mickaelistria:
As far as I know the plan is not ready yet, but it should be similar as https://www.eclipse.org/projects/project-plan.php?planurl=http://www.eclipse.org/eclipse/development/plans/eclipse_project_plan_4_26.xml ; but starting mid-end December to release 4.27 in March.

mickaelistria:
Closed with #118

PR #118, opened 2022-11-24
Add extension point for configuring ITextDoubleClickStrategy.

andrewL-avlq:
Provide extension point so that the double click strategy used by the Generic Editor can be configured by content type.

See https://github.com/eclipse-platform/eclipse.platform.text/issues/113


mickaelistria:
Can you please rebase on top of master? You'll probably have to add +0.0.100 to bundle version as we're starting a new Platform version.

mickaelistria:
Thank you!

PR #123, opened 2022-12-08
Keep the same order of install / uninstall than before.

angelozerr:
Keep the same order of install / uninstall than before.


angelozerr:
As it seems the order of install is very important for reconciler, content assisiet and presentation reconciler, the install / unisntall for those component doesn't use the generic list lifecycles.  This lifecycles list is used for new component like IAutoEditingtrategies.

cdietrich:
my test is green again.  will test if memory still is good over lunch

mickaelistria:
Thanks for the fix!

Issue #122, opened 2022-12-07
InlinedAnnotationSupport.findExistingAnnotation() performance issues with a lot of inlined annotations and a lot of positions


AlexOnisa:
Currently when using the latest version (3.13) of the InlinedAnnotationSupport class for a text viewer with a lot of positions that have AbstractInlinedAnnotation's, the findExistingAnnotation() method is poorly performing.
In our use case when having more than 100.000 annotations and 100.000 positions to verify it takes more than 15 minutes to do findExistingAnnotation(), and I think you know that it's obvious why, the current complexity is of O(n), as it iterates the whole Set to get to its position.
In the meantime as a workaround I've overriden the InlinedAnnotationSupport behaviour to keep the fInlinedAnnotations field into a map based on the Position, so that the findExistingAnnotation() complexity is O(1) all the time, and the performance problem is solved.
But I don't like it because I've used reflection to do so.

Personally, I see it as a relevant issue and very important to fix it as it would benefit the performance of this class by a lot and overall of the whole plugin.

vogella:
Could you provide a pull request for the change?

AlexOnisa:
Hi @vogella, I've provided a PR for it.

iloveeclipse:
@mickaelistria : please don't forget to push your revert https://github.com/eclipse-platform/eclipse.platform.text/commit/f15490c0f40c70ad08b3ae0a5a5d115afececb79

vogella:
@AlexOnisa can you have another fix for the problem?

mickaelistria:
Maybe using AnnotationModel.getAnnotationIterator() to lookup would perform better than having dedicated lookup in InlinedAnnotationSupport?

vogella:
Please re-open if you find a better solution

PR #124, opened 2022-12-08
#122

AlexOnisa:


org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
```diff
@@ -333,9 +337,10 @@ public void mouseUp(MouseEvent e) {
 	private AnnotationPainter fPainter;
 
 	/**
-	 * Holds the current inlined annotations.
+	 * Holds the current inlined annotations based on their {@link Position}s in a HashMap to make
+	 * lookup, insertion and deletion in average-case time complexity of O(1).
 	 */
-	private Set<AbstractInlinedAnnotation> fInlinedAnnotations;
+	private Map<Position, AbstractInlinedAnnotation> fInlinedAnnotationByPositionMap;
```
mickaelistria, line 343: There can be multiple annotation on the same Position, so it should be `Map<Position, Collection<AbstractInlinedAnnotation>>`

org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
```diff
@@ -545,8 +549,8 @@ private void removeInlinedAnnotations() {
 	 * @return the {@link AbstractInlinedAnnotation} from the given point and null otherwise.
 	 */
 	private AbstractInlinedAnnotation getInlinedAnnotationAtPoint(ISourceViewer viewer, int x, int y) {
-		if (fInlinedAnnotations != null) {
-			for (AbstractInlinedAnnotation ann : fInlinedAnnotations) {
+		if (fInlinedAnnotationByPositionMap != null) {
+			for (AbstractInlinedAnnotation ann : fInlinedAnnotationByPositionMap.values()) {
```
mickaelistria, line 553: Here it should interate on the map entries instead of values, and skip entries whose position isn't in the x/y range.

org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
```diff
@@ -485,16 +490,15 @@ public void updateAnnotations(Set<AbstractInlinedAnnotation> annotations) {
 	 */
 	@SuppressWarnings("unchecked")
 	public <T extends AbstractInlinedAnnotation> T findExistingAnnotation(Position pos) {
```
mickaelistria, line 492: And that makes me wonder whether this method is correct at all, even in the current/previous state: I don't think there is any guarantee that there is only 1 annotation for 1 position, is there?

AlexOnisa: From a user's point of view, in my opinion, it doesn't make sense to have multiple code annotations on the same range position or if the annotations are overlapping like so Pos1 = [0,1,2], Pos2 = [2,3,4].
I've created an example, I've modified the "InlinedAnnotationDemo" class example to show the overlapping of annotations:
img
Also to prove my point:
- check CodeMiningManager.renderCodeMinings() line 256, where you can clearly see that you can't have annotations on the same position, if you correctly use this mechanism and let the CodeMiningManager manage the annotations.
- in the InlinedAnnotationDemo.java you can also see that they tried to mimic the same behavior as above, where it verified if there is already an annotation for a position.
- the InlinedAnnotationSupport.updateAnnotations() method is called only from CodeMiningManager.renderCodeMinings() which guarantees that there is going to be just 1 annotation and 1 position.

I would say that if it's still not the case that it should be a 1:1 pair it should be made clear in which direction this code mining feature is going.
-Allow multiple annotations on the same position: then you would have to adapt the code and make it understandable that is possible.
-Not allow multiple annotations on the same position: then make it more restrictive if it's needed and document it.

org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
AlexOnisa: I do not want to modify the old logic here, the goal was only to improve findExistingAnnotation() and adapt all the places where it was affected. 
Here I just replaced some lines so that it works with the new data structure the same as before. 
You can freely make modifications, commit and have someone else more knowledgeable here to review the code.
Also I think that it's not enough to verify that the position is in the x/y range because there are additional verifications  based on the actual text height/width annotation that's in the UI, see AbstractInlinedAnnotation.contains(). Also I would introduce code duplication if I would move the code from there to getInlinedAnnotationAtPoint().

org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
mickaelistria: OK, so I believe the current code has a bug and a design flaw as there is no string reason to make an a position can only have 1 annotation. In practice, annotations can come from different providers and the case of identical positions is probable and not really hard to deal with and the API specification never excluded that.
But I get that your patch is an improvement over current state and that it doesn't exclude to fix the API later. So you can ignore that comment and we'll review the change as it without considering other pre-existing issue.

mickaelistria:
And actually, there are other issues with this whole API: the annotation shouldn't hold the position, as this position can change. The annotation model has all the smartness to get position for an annotation; so ideally, we'd just get rid of `getPosition()` and use the annotation model to get the position; this would most likely behave better overall.
Further patches to consolidate the implementation would be welcome ;)

mickaelistria:
Thanks for this fix!

org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationSupport.java
szarnekow: This map cannot work reliably.
Position is mutable on offset and length and the information is part of the hashCode. I suspect the annotations cannot be found anymore after some editing in the document happened. 

@mickaelistria please consider reverting this

mickaelistria: Thanks. I've checked in more details and the position that is part of the AbtractInlinedAnnotation seems to reference the same object than the one in the IAnnotationModel, which is updated on change. So it's not good to use it as Map key.
That kind of adds credit to my previous comment about storing positions in the IAnnotation being a bad idea in general and that instead this code should rely more on the IAnnotationModel when it comes to finding annotations or their position.

szarnekow: > Would you like to push the revert PR 

I don't that'd add value. I cannot seem to find the revert button here in the MR immediately and a local setup just to issue a git revert && git push + making a MR... no thanks.
Please just go ahead with the revert.

mickaelistria:
See https://github.com/eclipse-platform/eclipse.platform.text/pull/125

PR #127, opened 2022-12-15
Revert "Test fix: focus shell before focusing button"

iloveeclipse:
This reverts commit 01906e9811dcfde3a1c1a995230c4ef5dd0a216b as the original SWT change was reverted too via
https://github.com/eclipse-platform/eclipse.platform.swt/pull/503.


iloveeclipse:
The only fail is expected, because the corresponding SWT revert is not yet in the SDK we use for tests.

PR #132, opened 2023-01-06
Fix problem with empty code minings being shown

### LINKED ISSUE #131: Code Minings - Blank line when no occurences are found in Eclipse 2022-12
The following is taken from: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/369

Description

Hello there,

Since Eclipse 2022-12, whenever Code Minings is active and we check "Only if there is at least one result", the editor will still show a blank line when there is no occurence of a method reference. The same is true for reference of other kinds, like fields or types.

This worked fine on 2022-09.
Expected Outcome

I expect that if no reference is found for a given method, field or type, then no blank line is shown when "Only if there is at least one result" is checked.
Current Outcome

Currently, when "Only if there is at least one result" is not checked then a "virtual" line will be inserted in the Editor, above the method, type or field with the text "0 references". If however, that option is checked, it will insert a blank line above the method, type or field which is not referenced anywhere in the code.
Steps to reproduce

This can probably be seen very easily in just about any application, but if not I have this one:
I have prepared a sample application that has a very simple Payara Micro application
It can be found [here](https://github.com/CarlosMOGoncalves/eclipse-issue-tester)

This app does nothing special. It has a Jakarta REST endpoint but that is just for show.

    Fresh install Eclipse for Eclipse IDE for Enterprise Java and Web Developers 2022-12
    Checkout the project from my repository
    Activate Code Minings for every possible reference and check "Only if there is at least one result", like this:

image

    Then just check any class with methods tha have no reference, like NumericUtils.java, this will show up:

image

Final note

    sorry for all the Issue ceremony. I used a template from Payara issues as I think it is pleasant to work with.

Environment

    Distribution: Eclipse IDE for Enterprise Java and Web Developers 2022-12 2022-12 for Windows x64
    JDK Version: Corretto JDK 11
    Operating System: Windows 10

jjohnstn:
- fix CodeMiningLineHeaderAnnotation to add check for mining label being empty
- fixes #131

### DIFF
--- a/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningLineHeaderAnnotation.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningLineHeaderAnnotation.java
@@ -1,5 +1,5 @@
 /**
- *  Copyright (c) 2017 Angelo ZERR.
+ *  Copyright (c) 2017, 2022 Angelo ZERR.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


akurtakov:
@mickaelistria Please review.

mickaelistria:
thanks @jjohnstn

PR #135, opened 2023-01-26
Fix duplicate entries in QuickSearch dialog for nested projects

### LINKED ISSUE #31: Nested projects: Search and Quick search matches appear twice
Open one multi-module maven project, press Ctrl+Shift+Alt+L to invoke Quick Search, and search something. Every matching item will appear two times.
Or press Ctrl + H to invoke Search, in File Search tab and search something. Every match appears twice.

image

jjohnstn:
- add code to find innermost project file for a LineItem
- fixes #31

### DIFF
--- a/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/LineItem.java
+++ b/org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/LineItem.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013 Pivotal Software, Inc.
+ * Copyright (c) 2013, 2023 Pivotal Software, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -9,12 +9,16 @@
  *
  * Contributors:
  *    Pivotal Software, Inc. - initial API and implementation
+ *    Red Hat Inc. - fix duplicate items when projects are nested
  *******************************************************************************/
 package org.eclipse.text.quicksearch.internal.core;
 
 import java.util.Objects;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.search.internal.ui.text.FileMatch;
 
 @SuppressWarnings("restriction")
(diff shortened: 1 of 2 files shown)


jjohnstn:
@mickaelistria Please review.

laeubi:
This sound very promising! I wonder if this might be optimized by doing the following and only fallback to your search if the `IFile` is not a local files:

- Use `IFile#getLoaction().getFile()` and check it is not `null`
- Then use `IWorkspaceRoot.findFilesForLocation` to get a list of all items mapping to that physical file
- Choose the innermost project

It might even worth it to think about a new `IFile#getInnermostProject()` method, then you can even work on the internal state of the object what might be more efficient (or at laest can be optimized later on) and it would be usable on other places as well.

mickaelistria:
+1 with what @laeubi says, this is already implemented in multiple places with
```java
  URI locationUri = resource.getLocationURI();
  return locationUri == null ? resource : //
     Arrays.stream(resource.getWorkspace().getRoot().findFilesForLocationURI(locationUri)) //
				.sorted(Comparator.comparingInt(aFile -> aFile.getFullPath().segments().length)) //
				.findFirst() // shortest workspace (project reletive) full path means most nested project
				.orElse(resource);
```

> It might even worth it to think about a new IFile#getInnermostProject() method, then you can even work on the internal state of the object what might be more efficient (or at laest can be optimized later on) and it would be usable on other places as well.

The "innermost" is not related to the state of the file, but to the state of the workspace root, so such API would be `IWorkspaceRoot.findInnerMostFile(IFile)`. It would be welcome as this code is already repeated in multiple places.

szarnekow:
> .sorted(Comparator.comparingInt(aFile -> aFile.getFullPath().segments().length)) //

Even though there are not many locations (on avg), it might be worthwhile to consider using `min(Comparator)` rather than sorted.

mickaelistria:
> Even though there are not many locations (on avg), it might be worthwhile to consider using min(Comparator) rather than sorted.

Indeed, and thanks for noting! That's one more reason to have a good API for it, so we optimize it in one place instead of repeating less efficient constructs in multiple places ;)

laeubi:
> The "innermost" is not related to the state of the file, but to the state of the workspace root, so such API would be IWorkspaceRoot.findInnerMostFile(IFile). It would be welcome as this code is already repeated in multiple places.

Each file belongs to a Workspace that has a workspace root, so I would at least want a delegate in `IFile` so it is easier to find/use, to be honest, as a user I would exspect that `IFile#getProject` aslready return exactly that value, but thats nothing one can change I assume so `IFile#get<whateveritsnamed>Project()`  + a hint on `getProject` would make this more obvious and hopefully improve the overall userexperience over time.

merks:
Just an FYI.  People do bad things and some of those people are "us":

org.eclipse.e4.tools.emf.ui.internal.common.component.dialogs.ContributionDataFile

org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/LineItem.java
```diff
@@ -26,19 +30,50 @@ public class LineItem {
...
+			}
+		}
+		if (innermostProject != fileProject) {
+			IPath newFilePath = filePath.removeFirstSegments(currentPath.segmentCount());
+			IFile newFile = innermostProject.getFile(newFilePath);
+			if (newFile.exists()) {
+				file = newFile;
+			}
+		}
+		return file;
```
mickaelistria, line 74:
```suggestion  
		URI locationUri = resource.getLocationURI();
		return locationUri == null ? resource : //
			Arrays.stream(resource.getWorkspace().getRoot().findFilesForLocationURI(locationUri)) //
				.min(Comparator.comparingInt(aFile -> aFile.getFullPath().segments().length))  // shortest workspace (project reletive) full path means most nested project
				.orElse(resource);
```

mickaelistria (changes requested):

org.eclipse.text.quicksearch/src/org/eclipse/text/quicksearch/internal/core/LineItem.java
jjohnstn: Thanks Mickael.  I used your suggestion but had to modify the resource name to be file as passed in to the method.

mickaelistria (approved):

jabby:
Thank you all for your work. I know I will enjoy this improvement a lot.

PR #137, opened 2023-01-27
ReplaceRefactoring: improve performance

jukzi:
Get rid of O(n^2) compares during search&replace with hashing.

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/text/ReplaceRefactoring.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/text/ReplaceRefactoring.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2015 IBM Corporation and others.
+ * Copyright (c) 2007, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


org.eclipse.search/search/org/eclipse/search/internal/ui/text/ReplaceRefactoring.java
```diff
@@ -290,31 +290,36 @@ public boolean hasMatches() {
...
-				matches.add(match);
-				return false;
+		Object key; // IFileStore or URI, see URIUtil.equals(URI, URI)
+		try {
+			key = EFS.getStore(uri);
+		} catch (CoreException e) {
+			// fall back to default equality test
+			key = uri;
+		}
+		if (fAlreadyCollected.containsKey(key)) {
```
iloveeclipse, line 309: But that doesn't match what URIUtil does - it converts arguments before doing equals to `EFS.getStore(one)`. Here only the key is converted to IFileStore or URI, but so we compare apples with oranges?

jukzi: see line 304: i convert too. the map then stores both apples and oranges in the bin (line 322+323). So i can test if either apple or orange is inside.

iloveeclipse: Arrgh, I've overlooked that. 
Please add comment to the map definition that keys can be both apples and oranges.

iloveeclipse: Beside this, LGTM

jukzi: done

iloveeclipse (approved):
Thanks

PR #144, opened 2023-02-03
Add filtering support for duplicate files in nested projects

### LINKED ISSUE #143: Add filter to File Search to remove nested project duplicate files
When nested projects are searched, multiple matches will occur for each project that is above the matched file.  The file should be opened in the context of the innermost project and this is already done, but some users might want to pare down the list of matches to exclude the duplicates that don't occur in the innermost project.  This is an enhancement request to add a filter to do this in the Search result view.

jjohnstn:
- add filter to File Search to remove duplicate file entries that are attributed to different projects in a nested project scenario
- fixes #143

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.properties
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.properties
@@ -98,7 +98,7 @@ SearchPage_wholeWord= Who&le word
 TextSearchEngine_statusMessage= Problems encountered during text search.
 
 TextSearchInnermostProjectFilter_action_label=Show only most &nested match
-TextSearchInnermostProjectFilter_name= Nested project duplicates
+TextSearchInnermostProjectFilter_name=Duplicate match from outer project
 TextSearchInnermostProjectFilter_description= For files in nested projects, don't show matches for duplicate file references that are owned by projects that aren't the innermost project.
 TextSearchPage_searchIn_label=Search In
 TextSearchPage_searchDerived_label=Deri&ved resources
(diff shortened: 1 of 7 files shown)


jjohnstn:
Hi @mickaelistria please review.  This is based on @merks suggestion to avoid duplicates via a filter rather than in the search resuilts.

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.properties
```diff
@@ -97,6 +97,9 @@ SearchPage_wholeWord= Who&le word
 
 TextSearchEngine_statusMessage= Problems encountered during text search.
 
+TextSearchInnermostProjectFilter_action_label= &Nested project duplicates
```
mickaelistria, line 100:
```suggestion
TextSearchInnermostProjectFilter_action_label=Show only most &nested match
```
?

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.properties
```diff
@@ -97,6 +97,9 @@ SearchPage_wholeWord= Who&le word
...
+TextSearchInnermostProjectFilter_name= Nested project duplicates
```
mickaelistria, line 101:
```suggestion
TextSearchInnermostProjectFilter_name=Duplicate match from outer project
```
?

jjohnstn:
Thanks @mickaelistria I used your string commits (interesting feature of github).  I don't think the filtered match label stuff needs to be sectioned out as it is essentially copied from the Java search page and just modified to take into account if the limit is also reached.  As well, there weren't any filters being set up for the File page before.  This plus the quick search change might be noteworthy additions, but I'll let you decide.

PR #148, opened 2023-02-08
Move/copy/adapt deprecated NavigatorDragAdapter

mickaelistria:


akurtakov:
This doesn't seem used anywhere. Shouldn't it replace the old one in FileSearchPage?

mickaelistria:
> This doesn't seem used anywhere. Shouldn't it replace the old one in FileSearchPage?

Indeed, forgot a `git add`.

PR #156, opened 2023-03-07
Add updateRelease.yml

akurtakov:


akurtakov:
@laeubi Any idea what went wrong in https://github.com/akurtakov/eclipse.platform.text/actions/runs/4352345143/jobs/7605013788 ?

akurtakov:
I get what is wrong - it tries git add **/pom.xml but everything is pomless so updateRelease.yml needs a change.

laeubi:
@akurtakov the problem is that I don't have enough git-power to get something that includes ANY `pom.xml` (including the root!) so i used two path specs but it seems git for some reason don'T like that .... its a bit of a mystery to me ;-)

laeubi:
> I'll push the change as the fix to the action has to go in the action repo itself.

You can also use a forked "releng" repo for experiments with the "included" workflow :-)

akurtakov:
It is probably bailing on first issue and not even trying next.

laeubi:
Yes probabbly but we already have had sucessfull runs, so its a bit confusing (for me), so if you can suggest a more gracefull pattern that would be great!

PR #162, opened 2023-03-17
Update Jenkinsfile to build using CentOS 8

akurtakov:
Trying to get rid of CentOS 7 image building.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,7 @@ pipeline {
 		disableConcurrentBuilds(abortPrevious: true)
 	}
 	agent {
-		label "centos-7"
+		label "centos-8"
 	}
 	tools {
 		maven 'apache-maven-latest'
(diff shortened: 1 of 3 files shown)


iloveeclipse:
```
org.eclipse.jface.text.tests.contentassist.ContextInformationTest.testContextInfo_hide_focusOut  Time elapsed: 0.249 s  <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.eclipse.jface.text.tests.contentassist.ContextInformationTest.testContextInfo_hide_focusOut(ContextInformationTest.java:117)
```

One stable test fail. But I assume that is something one can fix / mute, since that is about focus handling, which is usually very sensitive.

PR #164, opened 2023-03-17
Run the build on CentOS 9

akurtakov:


mickaelistria:
Note that on centos-9, we could switch the tests to use wayland

akurtakov:
There would be no centos9 as per https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2827#note_1092972

akurtakov:
Maybe we can switch to ubuntu latest and try using wayland.

mickaelistria:
An alternative would be to define our own containers to use for CI, like it's done for wildwebdeveloper.

PR #168, opened 2023-03-20
Adjust ContextInformationTest#testContextInfo_hide_focusOut

laeubi:
- if regular focus request fails, try to force the focus
- if that also fails, give an error
- after that, wait if the desired condition will be reached

WDYT @akurtakov @iloveeclipse ?


iloveeclipse:
If that works, it is better as disabling the test completely.
Please note, build fails with ECA check.

org.eclipse.jface.text.tests/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.jface.text.tests
-Bundle-Version: 3.13.0.qualifier
+Bundle-Version: 3.13.1.qualifier
```
iloveeclipse, line 5: 100

iloveeclipse (changes requested):

iloveeclipse (approved):
LGTM

laeubi:
> If that works, it is better as disabling the test completely.

I'll first check that the build succeeds then I'll enable centos-8 ... this is just a guess as I have no centos to test with.

vogella:
Disablement of test and migration to latest Cent OS https://github.com/eclipse-platform/eclipse.platform.text/pull/162

PR #171, opened 2023-03-28
Fix ClassCastException in TextSearchVisitor.TextSearchJob

jjohnstn:
- TextSearchVisitor.TextSearchJob.processFile() has a catch for FileCharSequenceProvider.FileCharSequenceException and tries to rethrow the exception cause.  In the case where the cause is an IOException, throw the original FileCharSequenceException which is a RuntimeException
- discovered by https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/436

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -223,8 +223,8 @@ public class TextSearchVisitor {
 						}
 						occurences = locateMatches(file, charsequence, matcher, monitor);
 					} catch (FileCharSequenceProvider.FileCharSequenceException e) {
-						if (e.getCause() instanceof RuntimeException) {
-							throw (RuntimeException) e.getCause();
+						if (e.getCause() instanceof RuntimeException runtimeEx) {
+							throw runtimeEx;
 						}
 						throw e;
 					}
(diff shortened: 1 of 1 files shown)


jjohnstn:
@akurtakov Please review.

org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -223,7 +223,10 @@ public IStatus processFile(List<IFile> sameFiles, IProgressMonitor monitor) {
 						}
 						occurences = locateMatches(file, charsequence, matcher, monitor);
 					} catch (FileCharSequenceProvider.FileCharSequenceException e) {
-						throw (RuntimeException) e.getCause();
+						if (e.getCause() instanceof RuntimeException) {
+							throw (RuntimeException) e.getCause();
+						}
```
mickaelistria, line 228: Let's try Java 17 way:
```suggestion
						if (e.getCause() instanceof RuntimeException runtimeEx) {
							throw runtimeEx.getCause();
						}
```

jjohnstn: Weirdly, when I change to the Java 17 format, I get an error "Unhandled exception type Throwable".  Do you see this too?  With the previous version, no error.

merks: You don't want to call getCause on runtimeEx.  It's already thing you want to throw.  😬

jjohnstn:
Thanks @merks

Issue #177, opened 2023-04-05
Remove AnnotationRulerColumn code added for bug 531952


iloveeclipse:
I've stumbled over the ruler code once again, and see that the code in question that was added [for this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=531952) can be simply commented out without any visible regression (at least I don't see [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=531952) appearing again).

Let try this instead of introducing inconsistencies in the code.

_Originally posted by @iloveeclipse in https://github.com/eclipse-platform/eclipse.platform.text/issues/172#issuecomment-1497377869_

PR #178, opened 2023-04-05
Remove AnnotationRulerColumn code added for bug 531952

iloveeclipse:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=531952 for original issue.

The code (text change listener) that was added to AnnotationRulerColumn to fix bug 531952 can be simply commented out without any visible regression.

Let's remove this code, it adds unneeded complexity & is inconsistent with the rest of the text listeners code that is supposed to be used from UI thread only.

Fixes https://github.com/eclipse-platform/eclipse.platform.text/issues/177


iloveeclipse:
@angelozerr : Angelo, as you have originally reported https://bugs.eclipse.org/bugs/show_bug.cgi?id=531952, could you please check if that works in your environment / with your code too? 

At least on Linux / latest master state I don't see any regressions like described in your original bug.

iloveeclipse:
@angelozerr , @mickaelistria : ping.

mickaelistria (approved):
I think the VisibleLinesTracker, that takes care of redrawing annotation whenever 1 visible line has a different position or height, is good enough.

Issue #47, opened 2022-06-22
brackets highlighting should not contains code mining text 


humphreygao:
![image](https://user-images.githubusercontent.com/52394598/174938526-1e3aee64-e1af-40be-a9d5-c51f2f8e1a89.png)
please see the picture above, the caret is placed at the closing bracket, and the matching bracket highlighting contains the method parameter name codemining text.

arihunta:
Hi folks, I'm hopeful new Eclipse contributor, and started looking at this as a possible first issue to tackle (in the hopes that it would be easy 🙃). Did a bit of poking around at the code, will drop the relevant links here. I'm not 100% sure I'm making much sense, being unfamiliar with the codebase, and UI programming at this level in general...

Seems the key class is `org.eclipse.jface.text.source.MatchingCharacterPainter`, and the draw in that happens [here](https://github.com/eclipse-platform/eclipse.platform.text/blob/master/org.eclipse.jface.text/src/org/eclipse/jface/text/source/MatchingCharacterPainter.java#L254-L258).

That delegates to `org.eclipse.swt.custom.StyledText.getTextBounds(int, int)` and then eventually `org.eclipse.swt.graphics.TextLayout.getBoundsInPixels(int, int)` (I was in a `win32` source directory [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/TextLayout.java#L1688), guess I'd need to check other OSes as well...).

For the example I was looking at, `start` and `end` passed into that are the same. The length is based on a "run", whose length is determined [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/TextLayout.java#L3755-L3778) with some calls down to the OS. 

On the other side of things, the code mining annotation seems to be represented by `org.eclipse.jface.internal.text.codemining.CodeMiningLineContentAnnotation` [here](https://github.com/eclipse-platform/eclipse.platform.text/blob/master/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/codemining/CodeMiningLineContentAnnotation.java).

[...]

PR #181, opened 2023-04-16
brackets highlighting should not contains code mining text #47

arihunta:
### Notes

`MatchingCharacterPainter` is using the `StyledText.getTextBounds()` method to determine the width of the highlight box to draw. This method calculates width based on runs of text, and thus includes the method parameter hints in that width.

This change instead moves to using `GC.textExtent()` to determine the width of text to highlight, ensuring that only the desired character is highlighted. The character width to highlight was already hard-coded to `1`, so now we just pull a single character of text from the desired offset, and pass it to `GC.textExtent()` to get its width.

### Testing

**Manual testing**

Tested on Windows, running in a normal window

image

Tested on Windows, on a desktop with 200% font scaling enabled:

image

**Unit testing**

I didn't find existing unit tests for `MatchingCharacterPainter`, and am not quite sure how one would implement such, though I'm very open to suggestions.


mickaelistria:
Thanks for this good patch!

PR #183, opened 2023-04-24
Hide the popup in async mode when there are no completions left

rubenporras:
Fix bug a introduced in 78fc9f6f6566323c691f1fee1eb14de2f27e7d34 that prevented the popup from closing when there are were completions left.

In addition, create the variable stillComputing earlier so that it can be reused.


rubenporras:
@mickaelistria , this fixes https://github.com/eclipse/lsp4e/issues/550 for me. Would you like to do the review?

rubenporras:
Hi @mickaelistria , 

thanks for the review. Would it be possible to downbrach to the state that was released for 2022-12 or 2023-03 and do a fix release? This bug is very visible for our end users.

If that is not something that we do for this project, are there snapshots that I could add to a target file to get the fix before the simrel 2023-06 RC1 is created?

Regards

akurtakov:
Even if the patch is backported to old releases there are no builds done for old branches so you would have to build it yourself. 
For the current development branch there is https://download.eclipse.org/eclipse/updates/4.28-I-builds/ you can use.

PR #184, opened 2023-04-27
 unlimited searchs can accidently hang/crash the system #97 

r6squeegee:
this is to help address https://github.com/eclipse-platform/eclipse.platform.ui/issues/872

I should note though, I'm unsure of how the junit tests work here, as they don't resemble junit tests I'm familiar with and looking at the existing ones... I'm not so sure if or where the other preferences are tested. As a result I've tested this manually but have not added any junit tests.

I also have no idea how the translations texts works here, I found the english text and added those but that's it.

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
@@ -528,9 +528,14 @@ public class TextSearchVisitor {
 				ReusableMatchAccess access= new ReusableMatchAccess();
 				access.initialize(file, start, end - start, searchInput);
 				occurences.add(access);
-				boolean res= fCollector.acceptPatternMatch(access);
-				if (!res) {
-					return occurences; // no further reporting requested
+				try {
+					boolean res = fCollector.acceptPatternMatch(access);
+					if (!res) {
+						return occurences; // no further reporting requested
+					}
+				} catch (OperationCanceledException e) {
+					monitor.setCanceled(true);
+					return occurences;
 				}
 			}
 			// Periodically check for cancellation and quit working on the current file if the job has been cancelled.
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.java
@@ -163,6 +163,7 @@ public final class SearchMessages extends NLS {
 	public static String FileSearchQuery_singularLabel;
 	public static String FileSearchQuery_singularLabel_fileNameSearch;
 	public static String FileSearchQuery_pluralPattern_fileNameSearch;
+	public static String FileSearchQuery_maxResultsExceeded;
 	public static String OpenSearchDialogAction_label;
 	public static String OpenSearchDialogAction_tooltip;
 	public static String FileTypeEditor_typeDelimiter;
@@ -205,6 +206,7 @@ public final class SearchMessages extends NLS {
 	public static String SearchPreferencePage_defaultPerspective_none;
 	public static String SearchPreferencePage_ignorePotentialMatches;
 	public static String SearchPreferencePage_rememberLastUsedPage;
+	public static String SearchPreferencePage_maxResults;
 	public static String ReplaceAction_label_all;
 	public static String ReplaceAction_label_selected;
 	public static String SelectAllAction_label;
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
@@ -25,6 +25,7 @@ import org.eclipse.jface.preference.ColorFieldEditor;
 import org.eclipse.jface.preference.ComboFieldEditor;
 import org.eclipse.jface.preference.FieldEditorPreferencePage;
 import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.preference.IntegerFieldEditor;
 import org.eclipse.jface.preference.PreferenceConverter;
 import org.eclipse.jface.util.PropertyChangeEvent;
 
@@ -44,7 +45,6 @@ public class SearchPreferencePage extends FieldEditorPreferencePage implements I
 
 	public static final String PAGE_ID= "org.eclipse.search.preferences.SearchPreferencePage"; //$NON-NLS-1$
 
-
 	public static final String IGNORE_POTENTIAL_MATCHES= "org.eclipse.search.potentialMatch.ignore"; //$NON-NLS-1$
 	public static final String REMEMBER_LAST_USED_PAGE = "org.eclipse.search.potentialMatch.remember"; //$NON-NLS-1$
 	public static final String EMPHASIZE_POTENTIAL_MATCHES= "org.eclipse.search.potentialMatch.emphasize"; //$NON-NLS-1$
@@ -56,12 +56,13 @@ public class SearchPreferencePage extends FieldEditorPreferencePage implements I
 	public static final String TEXT_SEARCH_ENGINE = "org.eclipse.search.textSearchEngine"; //$NON-NLS-1$
 	public static final String TEXT_SEARCH_QUERY_PROVIDER = "org.eclipse.search.textSearchQueryProvider"; //$NON-NLS-1$
 	public static final String LIMIT_HISTORY= "org.eclipse.search.limitHistory"; //$NON-NLS-1$
+	public static final String MAX_RESULTS = "org.eclipse.search.maxResults"; //$NON-NLS-1$
 
 	private ColorFieldEditor fColorEditor;
 	private BooleanFieldEditor fEmphasizedCheckbox;
 	private BooleanFieldEditor fIgnorePotentialMatchesCheckbox;
 	private BooleanFieldEditor fRememberLastUsedPageCheckbox;
-
+	private IntegerFieldEditor fieldMaxResults;
 
 	private static class PerspectiveDescriptorComparator implements Comparator<IPerspectiveDescriptor> {
 		@Override
@@ -91,6 +92,9 @@ public class SearchPreferencePage extends FieldEditorPreferencePage implements I
 		store.setDefault(TEXT_SEARCH_ENGINE, ""); //default search engine is empty string //$NON-NLS-1$
 		store.setDefault(TEXT_SEARCH_QUERY_PROVIDER, ""); // default query provider is empty string  //$NON-NLS-1$
 		store.setDefault(LIMIT_HISTORY, 10);
+
+		// we will use 0 to mean disable maxResults limiting functionality
+		store.setDefault(MAX_RESULTS, 0);
 	}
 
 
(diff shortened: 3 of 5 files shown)


org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
```diff
@@ -56,12 +56,15 @@ public class SearchPreferencePage extends FieldEditorPreferencePage implements I
 	public static final String TEXT_SEARCH_ENGINE = "org.eclipse.search.textSearchEngine"; //$NON-NLS-1$
 	public static final String TEXT_SEARCH_QUERY_PROVIDER = "org.eclipse.search.textSearchQueryProvider"; //$NON-NLS-1$
 	public static final String LIMIT_HISTORY= "org.eclipse.search.limitHistory"; //$NON-NLS-1$
+	public static final String MAX_RESULTS = "org.eclipse.search.maxResults"; //$NON-NLS-1$
+
+	private static final int MAX_RESULTS_DEFAULT = 2000;
```
jukzi, line 61: I would prefer a far higher default. Like the count when OOME or lags start. At which number do you get trouble?

jukzi:
it's just impossible to find all results when searching for "a" in a big workspace. However a hint that search limit was reached would be nice
for example here:
image
example shows also that 100_000 results are still no issue.

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
iloveeclipse: Fully agree here, 2000 is too small. 100.000 **at least** would be OK I guess.

org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -528,9 +528,18 @@ private List<TextSearchMatchAccess> locateMatches(IFile file, CharSequence searc
...
+					boolean res = fCollector.acceptPatternMatch(access);
+					if (!res) {
+						return occurences; // no further reporting requested
+					}
+				} catch (CoreException e) {
+					if (e.getStatus() == Status.CANCEL_STATUS) {
+						monitor.setCanceled(true);
+						return occurences;
+					} else {
+						throw (e);
```
iloveeclipse, line 541: `throw (e)` : please remove `()` around e.

org.eclipse.search/search/org/eclipse/search/internal/core/text/TextSearchVisitor.java
```diff
@@ -528,9 +528,18 @@ private List<TextSearchMatchAccess> locateMatches(IFile file, CharSequence searc
...
-				boolean res= fCollector.acceptPatternMatch(access);
-				if (!res) {
-					return occurences; // no further reporting requested
+				try {
+					boolean res = fCollector.acceptPatternMatch(access);
+					if (!res) {
+						return occurences; // no further reporting requested
+					}
+				} catch (CoreException e) {
+					if (e.getStatus() == Status.CANCEL_STATUS) {
```
iloveeclipse, line 537: Better is to catch `OperationCanceledException` which is what probably should be thrown

iloveeclipse (changes requested):
See comment https://github.com/eclipse-platform/eclipse.platform.text/pull/184#issuecomment-1525546051. That must be done in any case.

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
r6squeegee: The amount depends on the users workstation. As mentioned on my original report, my system had completely failed by 500k, it had been completely unresponsive way before then like at about 250k. I had only gone up to 500k because I was trying to stop / close eclipse normally. In the end a kill -9 was required for me. For reference my cpu was an i5-8250u, 2gb of ram for eclipse jvm, and it was all on an ssd.

Note that the count / memory is only part of the problem. The code base I was working on was very small. It was only about 1000 files, 50k lines of code, total source size of about 2mb of text files. The problem was most likely the amplification cause by the wildcard search matching literally everything in every file many times. 

I think any default size beyond 100k would defeat the purpose of this feature as by default, it would no longer provide the circuit breaker functionality of stopping a run away query. In many cases, I can no longer stop a query even after it hits just the 10k's, the best I can do often is wait for it to finish.

So given this discussion, I'm no longer confident that what I've submitted is of any use to either you guys nor myself at this point. I'm not sure where to go from here.

r6squeegee:
I've set the default for this to disabled, so this in theory should not affect anyone unless they explicitly turn it on. With it turned off it will skip all calls to getMatchCount()  so this should help with that regression issue.

iloveeclipse (commented):
This looks better now, unfortunately I will be able to review this in the IDE next week only.

r6squeegee:
oh and to further continue the how quickly it crashes/hangs conversation...

cpu : i5-8250u
jvm ram : the default (-Xms256m -Xmx2048m)
ssd : yes...

2 projects in my workspace
- eclipse.platform
- eclipse.platform.text

scenario :
- ctrl-H -> File Search : containing text = *
- time till eclipse completely hangs / unresponsive ~0.5 seconds

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.java
```diff
@@ -163,6 +163,7 @@ private SearchMessages() {
 	public static String FileSearchQuery_singularLabel;
 	public static String FileSearchQuery_singularLabel_fileNameSearch;
 	public static String FileSearchQuery_pluralPattern_fileNameSearch;
+	public static String FileSearchQuery_maxResultsExceeded;
```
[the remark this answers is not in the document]
r6squeegee: I believe it's in SearchMessages.properties (next file in the list)

org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchQuery.java
```diff
@@ -261,6 +271,18 @@ public String getSearchString() {
 	}
 
 	public String getResultLabel(int nMatches) {
+		String resultLabel = getResultLabelWithOutLimitString(nMatches);
+
+		int maxResultAllowed = SearchPreferencePage.getMaxResults();
```
jukzi, line 276: reusue maxSearchResults

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
```diff
@@ -91,6 +92,9 @@ public static void initDefaults(IPreferenceStore store) {
 		store.setDefault(TEXT_SEARCH_ENGINE, ""); //default search engine is empty string //$NON-NLS-1$
 		store.setDefault(TEXT_SEARCH_QUERY_PROVIDER, ""); // default query provider is empty string  //$NON-NLS-1$
 		store.setDefault(LIMIT_HISTORY, 10);
+
+		// we will use 0 to mean disable maxResults limiting functionality
+		store.setDefault(MAX_RESULTS, 0);
```
jukzi, line 97: I would prefer a good default. 100_000 seems to be a good compromise when problems start at 200_000

jukzi:
@r6squeegee please (always) squash all commits of a PR to a single commit. Exceptions being version counter increase.

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchMessages.java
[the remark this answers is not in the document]
r6squeegee: missed in the prior commit/push, just pushed it up now

jukzi:
> * time till eclipse completely hangs / unresponsive ~0.5 seconds

@r6squeegee please investigate what is eating all the memory. maybe there is some sort of leak for regexp?

org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchQuery.java
r6squeegee: I'm not sure I can, it's in a different class that's static. I don't think I have a reference to the collector here. I could change the collector to be an instance variable of this class, not sure if that's better or worst than just asking for the value from preferences again or not. If performance is a concern, I could make the maxSearchResults an instance variable, seems less intrusive than making the collector one.

jukzi: my idea was that i would like to see the value that was actually used.

org.eclipse.search/search/org/eclipse/search/internal/ui/SearchPreferencePage.java
r6squeegee: It seems like some people have 1,000,000 starting point desires and given that's not my usage scenario... I can't really contribute to that argument discussion, so I think I need to let those people decide between themselves and this can be changed here when the decision is made by everyone.

There's also the performance regression problem some one pointed out. By disabling it by default it would at least make sure no one gets a regression.

It's not my preferred option as I think a small default will provide the best out-of-the-box user experience for other people; but given the disagreement and regression, I think this at least helps this move forward for now.

iloveeclipse: Note, the performance problem was fixed yesterday via #189.

iloveeclipse:
Also please rebase your PR on top of master (no merge commits please)

org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileSearchQuery.java
r6squeegee: okay, amended, the msg now shows the limit you've set.

r6squeegee:
> > * time till eclipse completely hangs / unresponsive ~0.5 seconds
> 
> @r6squeegee please investigate what is eating all the memory. maybe there is some sort of leak for regexp?

interesting, with the latest checkout (that I just rebased upon)  it no longer happens (was still happening on my original checkout from a couple of weeks ago). I guess some one solved it some where.

mickaelistria:
So let's close it as #189 seems to have fixed it.
The track followed here was interesting initially but quickly lost focus on the actual issue. Usually, the best approach in case of a hang is not to reduce the feature, but really to investigate what is blocked and why. Calling `jstack` a few times and looking at what happens in the main thread tells what's blocked.

iloveeclipse:
Why not have a limit, if it can be iseful for someone?

mickaelistria:
OK, I've re-read #186 and indeed, a limit can be useful independently of UI. So let's reopen.

jukzi:
@r6squeegee do you plan to continue to work on a limit?

jukzi:
@iloveeclipse still changes requested?

iloveeclipse:
> @iloveeclipse still changes requested?

Will check next week. Going into longer weekend now.

laeubi:
This repository was merged with https://github.com/eclipse-platform/eclipse.platform.ui if your pull-request is still relevant please rebase the code and push it to the new repository.

Issue #187, opened 2023-04-27
Text search freezes UI


iloveeclipse:
Follow up on https://github.com/eclipse-platform/eclipse.platform.text/issues/186

This seem to be a regression **after** 4.25.

I can't reproduce on 4.25 but I can reproduce on master / SDK-I20230426-1800-linux-gtk-x86_64.

In my workspace that contains most of SDK bundles as sources case-insensitive searching for "a" in all files freezes UI after roughly 1.000.000 matches and at 2%, staying in that state with 100% CPU load for minutes. 
I have to kill IDE because it become not responsible.

In my (other and bigger) workspace on 4.25 search takes few seconds to find 26.000.000 matches without any freeze.

In the jstacks I see `AbstractTextSearchResult.getMatchCount()` that is almost always on top

```
"main" #1 prio=6 os_prio=0 cpu=538026.91ms elapsed=20829.88s tid=0x00007ffff0028f70 nid=0x9676 runnable  [0x00007ffff7fb6000]
   java.lang.Thread.State: RUNNABLE
	at java.util.concurrent.ConcurrentHashMap$Traverser.advance(java.base@17.0.6-internal/ConcurrentHashMap.java:3383)
	at java.util.concurrent.ConcurrentHashMap$ValueIterator.next(java.base@17.0.6-internal/ConcurrentHashMap.java:3483)
	at org.eclipse.search.ui.text.AbstractTextSearchResult.getMatchCount(AbstractTextSearchResult.java:305)
	at org.eclipse.search.internal.ui.text.FileTreeContentProvider.getMatchCount(FileTreeContentProvider.java:189)
	at org.eclipse.search.internal.ui.text.FileTreeContentProvider.elementsChanged(FileTreeContentProvider.java:239)
	- locked <0x0000000456e67178> (a org.eclipse.search.internal.ui.text.FileTreeContentProvider)
	at org.eclipse.search.internal.ui.text.FileSearchPage.elementsChanged(FileSearchPage.java:343)
	at org.eclipse.search.ui.text.AbstractTextSearchViewPage.runBatchedUpdates(AbstractTextSearchViewPage.java:1263)
	at org.eclipse.search.ui.text.AbstractTextSearchViewPage$UpdateUIJob.runInUIThread(AbstractTextSearchViewPage.java:154)
	at org.eclipse.ui.progress.UIJob.lambda$0(UIJob.java:148)
[...]

PR #188, opened 2023-04-27
Fix UI freeze on higher search matches count

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.text/issues/187

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileTreeContentProvider.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileTreeContentProvider.java
@@ -177,16 +177,19 @@ public class FileTreeContentProvider implements ITreeContentProvider, IFileSearc
 		if (element instanceof LineElement) {
 			LineElement lineElement= (LineElement) element;
 			IResource resource = lineElement.getParent();
-			if (getMatchCount(resource) > 0) {
+			if (resourceHasMatches(resource)) {
 				return lineElement.hasMatches(fResult);
 			}
 		}
 		return fPage.getDisplayedMatchCount(element) > 0;
 	}
 
-	private int getMatchCount(Object element) {
-		return fResult.getActiveMatchFilters().length > 0 ? fPage.getDisplayedMatchCount(element)
-				: fResult.getMatchCount();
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@jjohnstn : please review.

org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileTreeContentProvider.java
```diff
@@ -177,16 +177,19 @@ private boolean hasMatches(Object element) {
...
 		}
 		return fPage.getDisplayedMatchCount(element) > 0;
 	}
 
-	private int getMatchCount(Object element) {
-		return fResult.getActiveMatchFilters().length > 0 ? fPage.getDisplayedMatchCount(element)
-				: fResult.getMatchCount();
+	private boolean resourceHasMatches(IResource element) {
+		if (fResult.getActiveMatchFilters().length > 0) {
+			return fPage.getDisplayedMatchCount(element) > 0;
```
iloveeclipse, line 189: Note: I haven't debugged and checked if this branch need similar fix.
`fPage.getDisplayedMatchCount(element) > 0` is called twice and could be replaced with `fPage.hasMatches(element)` that could be implemented in a way that doesn't require calculating everything.

jjohnstn (commented):
@iloveeclipse I thought of adding the check for the filter but came up with a different solution.  Your solution solves the problem when the filter isn't specified but it will still cause a problem when filtering is specified.  I put the filtering logic up into the statement:

```
	@Override
	public synchronized void elementsChanged(Object[] updatedElements) {
		boolean singleElement = updatedElements.length == 1;
		Set<LineElement> lineMatches = Arrays.stream(updatedElements).filter(LineElement.class::isInstance)
				// only for distinct files:
				.map(u -> ((LineElement) u).getParent()).distinct()
				// query matches:
				.map(fResult::getMatchSet).flatMap(FileTreeContentProvider::toStream)
				.map(m -> ((FileMatch) m)).filter(this::isUnfiltered).map(m -> m.getLineElement())
				.collect(Collectors.toSet());

```

This no longer requires a check in the loop because the line elements have already been filtered from the lineMatches set.  What do you think?

PR #189, opened 2023-04-27
Fix UI freeze when searching

### LINKED ISSUE #187: Text search freezes UI

jjohnstn:
- add filtering to FileTreeContentProvider.elementsChanged() method so extraneous line elements are already filtered out of lineMatches set
- fixes #187

### DIFF
--- a/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileTreeContentProvider.java
+++ b/org.eclipse.search/search/org/eclipse/search/internal/ui/text/FileTreeContentProvider.java
@@ -18,6 +18,7 @@
 package org.eclipse.search.internal.ui.text;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -225,13 +226,26 @@ public class FileTreeContentProvider implements ITreeContentProvider, IFileSearc
 	@Override
 	public synchronized void elementsChanged(Object[] updatedElements) {
 		boolean singleElement = updatedElements.length == 1;
-		Set<LineElement> lineMatches = Arrays.stream(updatedElements).filter(LineElement.class::isInstance)
+		Set<LineElement> lineMatches = Collections.emptySet();
+		// if we have active match filters, we should only use non-filtered FileMatch
+		// objects
+		// to collect LineElements to update
+		if (fResult.getActiveMatchFilters().length > 0) {
+			lineMatches = Arrays.stream(updatedElements).filter(LineElement.class::isInstance)
 				// only for distinct files:
 				.map(u -> ((LineElement) u).getParent()).distinct()
 				// query matches:
 				.map(fResult::getMatchSet).flatMap(FileTreeContentProvider::toStream)
 				.map(m -> ((FileMatch) m)).filter(this::isUnfiltered).map(m -> m.getLineElement())
 				.collect(Collectors.toSet());
+		} else {
+			lineMatches = Arrays.stream(updatedElements).filter(LineElement.class::isInstance)
+					// only for distinct files:
+					.map(u -> ((LineElement) u).getParent()).distinct()
+					// query matches:
+					.map(fResult::getMatchSet).flatMap(FileTreeContentProvider::toStream)
+					.map(m -> ((FileMatch) m).getLineElement()).collect(Collectors.toSet());
+		}
 		try {
 			for (Object updatedElement : updatedElements) {
 				if (!(updatedElement instanceof LineElement)) {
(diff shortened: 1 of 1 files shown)


jjohnstn:
@iloveeclipse This is the other solution

iloveeclipse:
Jeff, I honestly have no idea what filters should filter in the patch (may be a comment in the code would help) and additionally I have no IDE access till weekend, so can't even debug or verify anything.

jjohnstn:
@iloveeclipse I'll add a comment.  The filter is set up to eliminate the same match being reported to two projects (one being nested in the other).  The filter takes a FileMatch input and decides if it is being reported on the innermost project.  The stream I am changing already is taking FileMatches and getting the line matches.  In the loop, I was eventually seeing if the line match should be filtered because it's matched file is an outer duplicate.  The count isn't relevant, it is whether the reported resource matches should be filtered.  I have moved the check up so the line matches are already cleared.  No count required and the filtering check is quick at this point.  I could make it even quicker by checking for the filter length and doing one of two calculations of the line matches where one does filter checking and the other doesn't.

I have verified that my solution continues to filter out duplicates in the file search results.  I'm sure yours will do the same.

Will either solution be able to be put into 4.28 M2?  I can't

iloveeclipse:
I would favor yours, since you know what the code is doing, and I just had some minutes between meetings to look at it :)
Please add comments to the code and I can merge.

jjohnstn:
@iloveeclipse Let me know if the comment is ok.  I added the check for active filters so the logic behaves exactly as it did before and should match or exceed the performance of #188.

iloveeclipse:
Jeff, the whole elementsChanged() code wasn't easy to understand and filtering added on top didn't make it easier. 
Your comment before on this PR explained things better - xould you please add something lile this to the javadoc of elementsChanged()?

iloveeclipse:
Thanks Jeff!

PR #190, opened 2023-05-03
Shutdown ForkJoinPool after code completion is executed

Christopher-Hermann:
The created ForkJoinPool must be shutdown after the asynchronous code completion is completed. Otherwise every code completion will create a new pool and they will never be shutdown.

This is a fix for a bug introduced with #185

### DIFF
--- a/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
@@ -374,19 +374,19 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 			return Collections.emptyList();
 		}
 		List<CompletableFuture<List<ICompletionProposal>>> futures = new ArrayList<>(processors.size());
-		for (IContentAssistProcessor processor : processors) {
-			// Use a custom ForkJoinWorkerThreadFactory, to prevent issues with a
-			// potential SecurityManager. Threads created by ForkJoinPool.commonPool(),
-			// which is used in CompletableFuture.supplyAsync(), get no permissions.
-			ForkJoinPool commonPool= new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
-					pool -> new ForkJoinWorkerThread(pool) {
-						// anonymous subclass to access protected constructor
-					}, null, false);
-			try {
+		// Use a custom ForkJoinWorkerThreadFactory, to prevent issues with a
+		// potential SecurityManager. Threads created by ForkJoinPool.commonPool(),
+		// which is used in CompletableFuture.supplyAsync(), get no permissions.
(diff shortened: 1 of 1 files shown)


Christopher-Hermann:
@sratz, @merks could you please review?

org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -382,7 +382,11 @@ protected List<CompletableFuture<List<ICompletionProposal>>> buildCompletionFutu
 					pool -> new ForkJoinWorkerThread(pool) {
 						// anonymous subclass to access protected constructor
 					}, null, false);
-			futures.add(CompletableFuture.supplyAsync(() -> this.getCompletionProposals(processor, invocationOffset), commonPool));
+			try {
```
iloveeclipse, line 385: Does it make sense to create and shutdown thread pool **per proposal**?

That should be done once for all!

merks:
Yes, why not create one pool shared for the show loop that's shut down once.

org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -374,19 +374,19 @@ protected List<CompletableFuture<List<ICompletionProposal>>> buildCompletionFutu
...
-			// which is used in CompletableFuture.supplyAsync(), get no permissions.
-			ForkJoinPool commonPool= new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
-					pool -> new ForkJoinWorkerThread(pool) {
-						// anonymous subclass to access protected constructor
-					}, null, false);
-			try {
+		// Use a custom ForkJoinWorkerThreadFactory, to prevent issues with a
+		// potential SecurityManager. Threads created by ForkJoinPool.commonPool(),
+		// which is used in CompletableFuture.supplyAsync(), get no permissions.
+		ForkJoinPool commonPool= new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
```
szarnekow, line 380: This still makes a new pool bound to the `invocationOffset`, e.g., for one invocation of content assist.

Please see this remark, which holds true here, too: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1017#issuecomment-1534323526

Christopher-Hermann: This seems to be a bigger change. I tend to revert #185 to prevent the creation of new pools for every code completion and maybe we find a proper solution for the issue.

iloveeclipse: > This seems to be a bigger change. I tend to revert #185 to prevent the creation of new pools for every code completion and maybe we find a proper solution for the issue.

Yes, please.

iloveeclipse:
Is this now abandoned? If yes, please close.

PR #195, opened 2023-05-04
Prevent multiple "Press F2 for focus" in Generic Editor

mickaelistria:
Fixes https://github.com/redhat-developer/eclipseide-jdtls/issues/42

### DIFF
--- a/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControl.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControl.java
@@ -114,7 +114,7 @@ public abstract class AbstractInformationControl implements IInformationControl,
 	 * @param statusFieldText the text to be used in the status field or <code>null</code> to hide the status field
 	 */
 	public AbstractInformationControl(Shell parentShell, String statusFieldText) {
-		this(parentShell, SWT.TOOL | SWT.ON_TOP, statusFieldText, null);
+		this(parentShell, SWT.TOOL | SWT.ON_TOP | SWT.RESIZE, statusFieldText, null);
 	}
 
 	/**
--- a/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/hover/CompositeInformationControl.java
+++ b/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/hover/CompositeInformationControl.java
@@ -36,15 +36,16 @@ import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.editors.text.EditorsUI;
 import org.eclipse.ui.internal.genericeditor.GenericEditorPlugin;
 
 public class CompositeInformationControl extends AbstractInformationControl implements IInformationControlExtension2 {
 
 	final LinkedHashMap<ITextHover, IInformationControlCreator> creators;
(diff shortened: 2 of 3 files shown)


org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControl.java
```diff
@@ -114,7 +114,7 @@ public abstract class AbstractInformationControl implements IInformationControl,
 	 * @param statusFieldText the text to be used in the status field or <code>null</code> to hide the status field
 	 */
 	public AbstractInformationControl(Shell parentShell, String statusFieldText) {
-		this(parentShell, SWT.TOOL | SWT.ON_TOP, statusFieldText, null);
+		this(parentShell, SWT.TOOL | SWT.ON_TOP | SWT.RESIZE, statusFieldText, null);
```
szarnekow, line 117: This looks like a pretty random change given that it didn’t change in the last 16 years. Can you clarify why it was necessary? Also I’m repeating my question why there is literally no time for any reviews before you merge your changes into the platform repository, @mickaelistria

iloveeclipse:
Mickael, neither the commit nor pr no issue explain the problem or solution. 
No test is added to show the problem / fix. The change is definitely not trivial. 
This is not OK to push something like this, especially for experienced contributors.

szarnekow:
Can you please answer https://github.com/eclipse-platform/eclipse.platform.text/pull/195#discussion_r1185127187 too, @mickaelistria

org.eclipse.jface.text/src/org/eclipse/jface/text/AbstractInformationControl.java
mickaelistria: This change to the default style is to allow a popup to have both status text and allow it to be resizable, also for consistency with other constructor (that takes a ToolbarManager as parameter) and overall better behavior in most cases. To achieve that, alternatives were either changing the default style for this constructor -which is the one used by the Generic Editor among other-, or un-deprecating the other constructor and extracting public constants for some default style, or extracting a method that creates the Shell so it can be overridden... Overall changing the default style seemed the simplest and best approach with a very low risk of it (making popup resizable) being perceived as a regression.
However, if one can suggest another way to have an AbstractInformationControl allowing a statusText and being resizable at the same time without changing the API or default style, I'd likely try it instead.

As for why I merge such change in a short time, it's simply because I find it much more efficient. I find that for changes that are relatively safe (ie not too error-prone, no API change, not scattered across several bundles... ie easy to revert),  efficiency is best by merging and maybe revert or improve later if need be; it's iterative development.
Committers are allowed to do so as per the Eclipse Development Process.

In this particular case, I now realize I should have extracted the SWT.RESIZABLE change into a dedicated commit with a dedicated message to clarify that particular bit of the change.

szarnekow: How does the change interact with the field `fResizable`? Why is that field even there if all visible constructors set the resize bit? Can and should other code be removed? Is it even safe to assume that all there controls should be resizable if there is special logic for resizable and non-resizable?

These are the first questions that immediately pop into my mind. My conclusion is: This change is not safe and will cause regressions for some products. Otherwise the code would not have been written like that. Please revert, make a ticket and cross-check with others whether the behavioral change is good or not.

mickaelistria: I will submit another patch to address the problem differently and remove the SWT.RESIZE default.

mickaelistria: See https://github.com/eclipse-platform/eclipse.platform.text/pull/197

PR #197, opened 2023-05-05
AbstractInformationControl constructor for resizable + status text

mickaelistria:
and use this constructor in Generic Editor which needs this combination

This also removes the addition of SWT.RESIZE style on existing constructor while this constructor specifies that the control is not resizable.
The new constructor takes the resizable as parameter.


szarnekow (approved):

mickaelistria:
Thanks for the review.

PR #205, opened 2023-05-19
Show the popup for auto activated only if there are completions to shown

rubenporras:
The fix improves the popup so that if it is autoactivated it is only shown if there are completions to be shown. Until know it could happen that the popup would be show with the message "Still computing (0%)".

This happens because there is a race condition between the PopupVisibleTimer and the lambdas executed in computeAndPopulateProposals. The PopupVisibleTimer is started in the async mode (that is, the servers did not respond in less than 50ms) and has a timeout of 500ms. That means there is a range between the 50 and the 500 ms that the server can respond with an empty completion list (in which case the lambdas executed in computeAndPopulateProposals will not make the popup visible). Then after 500ms, the PopupVisibleTimer will make the popup visible, and since all completable futures have been processed, the popup will stay open until the user closes it.

### DIFF
--- a/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
@@ -244,7 +244,7 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 						if (offset != fInvocationOffset || fComputedProposals != requestSpecificProposals) {
 							return;
 						}
-						boolean stillComputing= fComputedProposals.contains(computingProposal);
+						boolean stillComputing= stillComputing();
 						if (autoInsert
 								&& !autoActivated
 								&& !stillComputing
@@ -260,10 +260,7 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 						if (!stillComputing && callback != null) {
 							callback.accept(fComputedProposals);
 						} else {
-							boolean hasProposals= (stillComputing && fComputedProposals.size() > 1)
-									|| (!stillComputing && !fComputedProposals.isEmpty());
-
-							if ((autoActivated && hasProposals) || !autoActivated) {
+							if ((autoActivated && hasProposals()) || !autoActivated) {
 								setProposals(fComputedProposals, false);
 								displayProposals(true);
 							} else if (isValid(fProposalShell) && fProposalShell.isVisible() && remaining.get() == 0) {
@@ -406,6 +403,15 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 		return IDocument.DEFAULT_CONTENT_TYPE;
 	}
 
+	private boolean stillComputing() {
+		return fComputedProposals.stream().anyMatch(ComputingProposal.class::isInstance);
+	}
+
+	private boolean hasProposals() {
+		boolean stillComputing= stillComputing();
+		return (stillComputing && fComputedProposals.size() > 1) || (!stillComputing && !fComputedProposals.isEmpty());
+	}
+
 	private class PopupVisibleTimer implements Runnable {
 		private Thread fThread;
 
@@ -427,7 +433,11 @@ class AsyncCompletionProposalPopup extends CompletionProposalPopup {
 							fMutex.wait(fAutoActivationDelay);
 					}
 					Optional<Display> display= Optional.ofNullable(fContentAssistSubjectControlAdapter.getControl()).map(Control::getDisplay);
-					display.ifPresent(d -> d.asyncExec(() -> displayProposals(true)));
+					display.ifPresent(d -> d.asyncExec(() -> {
+						if (hasProposals()) {
+							displayProposals(true);
+						}
+					}));
 					break;
 				}
 			} catch (InterruptedException e) {


rubenporras:
@mickaelistria , sorry to ask, but this is new to me. What does it mean that "today is a freeze day"?

Is it possible still to fix bugs at this point in https://wiki.eclipse.org/Category:SimRel-2023-06?

As information if the bug is triggered, since the pop up steals the focus, the user is forced to press ESC, or arrow left/right (or something with the mouse). Arrow up/down does not work because the popup only has one entry. It was happening quite reliable for me while testing our implementation based on LSP4E and it was very annoying.

Regards

mickaelistria:
Eclipse Platform has "freeze" days when building milestones or releases and during those freeze days, it's not permitted to merge code in. Moreover, Platform is currently in RC phase, so only regression fixes or trivial fixes are admitted. I don't think this patch qualifies as any of those. This will have to wait for the next version (4.29 / 2023-09).

rubenporras:
Hi @mickaelistria ,

it is a regression introduced in https://github.com/eclipse-platform/eclipse.platform.text/commit/78fc9f6f6566323c691f1fee1eb14de2f27e7d34, where the timer was introduced. I can also simplify the PR to do less cleanup if you like so that it is only changing the run() method, than change would be very local.

Since the last released LSP4E forces an upgrade on the eclipse platform, we cannot avoid this regression.

Regards

rubenporras:
Thanks, I have done that. I am happy to make the change even less intrusive by using the new functions only in the `run` method, and keep the other method exactly as it is now. Then I would refactor the code to make it nice once the 2023-06 is out.

mickaelistria:
Just for completeness, can you please describe here some steps to reproduce the issue (ideally with an editor from the Eclipse SDK (I believe editing a .target file with Generic editor is a good candidate), and validate this patch fixes it?

org.eclipse.jface.text/src/org/eclipse/jface/text/contentassist/AsyncCompletionProposalPopup.java
```diff
@@ -427,7 +433,11 @@ public void run() {
 							fMutex.wait(fAutoActivationDelay);
 					}
 					Optional<Display> display= Optional.ofNullable(fContentAssistSubjectControlAdapter.getControl()).map(Control::getDisplay);
-					display.ifPresent(d -> d.asyncExec(() -> displayProposals(true)));
+					display.ifPresent(d -> d.asyncExec(() -> {
+						if (hasProposals()) {
+							displayProposals(true);
```
mickaelistria, line 438: Do I get it right that this wouldn't show the popup while proposals are computing? IMO, it's interesting to show the "Computing" feedback to users.
I think the condition should be `if (stillComputing() || hasProposls())` and then, if there are no proposals `if (!stillComputing() && !hasProposals() && !isAutoActivated())` then hide the popup.

rubenporras: Yes, I agree that if the computing has been requested, the feedback can be shown to the user. I will update the PR to avoid the popup only if is autoactivated.

rubenporras: @mickaelistria , I have checked the code again. The condition `!isAutoActivated()` is implicit in the `run` method, because the `PopupVisibleTimer` is only started if `!autoActivated`, this is show in the excerpt:

```

		displayProposals(!autoActivated);
	}

	@Override
	void displayProposals(boolean showPopup) {
		if (showPopup) {
			fPopupVisibleTimer.stop();
		}

		super.displayProposals(showPopup);
		if (!showPopup) {
			fPopupVisibleTimer.start();
		}
	}
```

So, I think the PR is fine from this point of view. I will write a second comment on the other review item: if the computing has been requested, the feedback can be shown to the user. I knew that when I coded the PR, but forgot over the weekend about it :)

rubenporras: > Do I get it right that this wouldn't show the popup while proposals are computing? IMO, it's interesting to show the "Computing" feedback to users. I think the condition should be `if (stillComputing() || hasProposls())` and then, if there are no proposals `if (!stillComputing() && !hasProposals() && !isAutoActivated())` then hide the popup.

After looking at this again, I think one can actually remove the `PopupVisibleTimer` that was introduced in https://github.com/eclipse-platform/eclipse.platform.text/commit/78fc9f6f6566323c691f1fee1eb14de2f27e7d34. By the comment on that commit:

> if async completion is triggered automatically, then the popup is only shown if there are completions to be shown

but that is not what happens if the `PopupVisibleTimer`, the timer as introduced in that commit, will just show the popup after the `fAutoActivationDelay` of 500 passes. The code in the lambda `computeAndPopulateProposals` for the async case will already show the popup the first time there is a completion proposal added, and it will also close it if it is open and the last answer arrives.

My PR adds the same condition in the timer as in the `computeAndPopulateProposals`, so I think it is correct as it is, but we could as well remove the timer, or am I overlooking something. @gayanper , @vogella , you worked on the commit that introduced the problem. Do you have any opinion on my comment that we could remove the timer or if we should apply the condition as in this PR or as suggested by @mickaelistria?

Thanks

rubenporras:
> Just for completeness, can you please describe here some steps to reproduce the issue (ideally with an editor from the Eclipse SDK (I believe editing a .target file with Generic editor is a good candidate), and validate this patch fixes it?

Regarding this, it is a race condition between the `PopupVisibleTimer` and the lambdas executed in `computeAndPopulateProposals`. The `PopupVisibleTimer` is started in the async mode (that is, the servers did not respond in less than 50ms) and has a timeout of 500ms. That means there is a range between the 50 and the 500 ms that the server can respond with an empty completion list (in which case the lambdas executed in `computeAndPopulateProposals` will not make the popup visible). Then after 500ms, the `PopupVisibleTimer` will make the popup visible, and since all completable futures have been processed, the popup will stay open until the user closes it.

I will add this to the description of the PR.

mickaelistria:
I think the fact that we're not even sure whether it's better to revert part of previous commit or whether it's better to tweak existing code is a sign that the issue is not trivial to resolve. Also, this is not fully a bugfix but a change in the functional behavior (we replace showing feedback of emptiness by no feedback), For those reasons, I think it's a bit risky to do in in RC and believe it's worth delaying it for 4.29.

iloveeclipse:
> I think it's a bit risky to do in in RC and believe it's worth delaying it for 4.29.

Yes. Please don't merge in 4.28.

rubenporras:
Is the proposal to revert https://github.com/eclipse-platform/eclipse.platform.text/commit/be9ac163f257f88e0f5a64c4ed93bde68207bf34 and then https://github.com/eclipse-platform/eclipse.platform.text/commit/78fc9f6f6566323c691f1fee1eb14de2f27e7d34?

that would be fine for me. Reverting only my last commit, would be a bad idea, it would leave the editor unusable for anyone having a language server which cannot respond in 50 ms. It would be unusable because each time the user stops typing, after 50 ms he would see a popup that never closes and that steals the focus. If at the point where he stopped writing there are no possible completions he will just see a message "Computing proposals" forever, which is even worse. The user did not asked for the feedback, he was just typing.

That is why I think that we need to revert both or fix the problem.

mickaelistria:
> That is why I think that we need to revert both or fix the problem.

I think none is doable safely and properly in RC.

rubenporras:
@mickaelistria, I think we can do a small fix that should improve the situation quite a lot (see https://github.com/eclipse-platform/eclipse.platform.text/pull/208 which I have created now). It is like a revert of my previous PR and a very small addition to fix the problem.

I understand that we did RC1 now and that changes should be very local (which I try to do), but on the other hand, I think we should fix clear bugs that we discover if they can be fixed locally. Otherwise we accept to deliver something in bad state and I would not understand what is the point of an RC where one cannot fix regressions.

mickaelistria:
@rubenporras Can you please update this one into a proposal for 4.29 M1 since we merged #208 for 4.28 RC1 ?

rubenporras:
Yes, I will do that.

PR #207, opened 2023-05-23
Revert "Hide the popup in async mode when there are no completions left"

rubenporras:
Reverts eclipse-platform/eclipse.platform.text#183, as it causes the PopupVisibleTimer to be kept running when the last remaining proposal computation finishes.


rubenporras:
@mickaelistria , this reverts my last PR.

PR #208, opened 2023-05-23
Reverts #183, as it causes the PopupVisibleTimer to be kept running when

rubenporras:
the last remaining proposal computation finishes.

Instead the tear down is done either when it was not visible (as before #183) or when it was visible but there is no proposals to show.


mickaelistria:
@vogella @gayanper Can one of you please review this for RC1?

PR #214, opened 2023-05-26
Use less aggressive defaults

rubenporras:
Both the JDT and VSCode do not trigger auto completion on each alphanumeric character by default. They only do it on trigger characters and only with a delay of 10 milliseconds.

This PR change the configuration of the Generic Editor to have the same defaults

### DIFF
--- a/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
+++ b/org.eclipse.ui.genericeditor/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.ui.genericeditor;singleton:=true
-Bundle-Version: 1.3.0.qualifier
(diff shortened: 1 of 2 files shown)


rubenporras:
I do not know if this default would make other people happy and would remove the need to have user/integration configuration options, but it would definitely for us.

I guess such a change cannot go after RC1, even if we would agree that it is a better default, but if it can go, it would be very good for us.

mickaelistria:
I think a small delay makes sense, we can maybe even push to 20ms as someone who types that fast is really typing "automatically" and most likely not interested in potential alternatives to what they're willing to type.
However, I personally like the fact that we have autocomplete on every char and miss it a lot when I use JDT or VSCode. I'd prefer it to remain turned on by default.

rubenporras:
I have updated the PR to only adapt the activation delay. I can also use 20 if you really prefer it, or 10 as in VSCode, I would not mind either.

mickaelistria (approved):

rubenporras:
I have bumped the version number

mickaelistria:
Thank you!

PR #224, opened 2023-06-12
Use try-with-resource and java.io.InputStream convenience methods

jukzi:
Doesn't try to fix anything. Just code and warnings reduced.


jukzi:
@trancexpress looks like we have 4 times almost the same code to read streams skipping BOM just to OOME check how the result is used.

PR #227, opened 2023-06-14
remove redundant Math.round(int)

jukzi:


org.eclipse.ui.examples.javaeditor/Eclipse Java Editor Example/org/eclipse/ui/examples/javaeditor/JavaContentOutlinePage.java
```diff
@@ -76,7 +76,7 @@ protected class ContentProvider implements ITreeContentProvider {
 		protected void parse(IDocument document) {
 
 			int lines= document.getNumberOfLines();
-			int increment= Math.max(Math.round(lines / 10), 10);
+			int increment= Math.max(lines / 10, 10);
```
mickaelistria, line 79: I imagine the intended behavior is more `Math.max(Math.round(lines / 10.), 10)` so in case of 109 lines , the increment is 11 and not 10.

jukzi: May be.  But the constant "10" seems already be arbitrary, so that it is also just arbitrary to round in any direction. So i decided to just not change the result.

mickaelistria: OK

Issue #228, opened 2023-06-16
[GenericEditor] An NPE occurs in `IndentFoldingStrategy.reconcile`


vrubezhny:
When running WWD JUnit tests on MacOS Github Actions runner a `NullPointerException` gets logged:

```
##[error]Exception in thread "org.eclipse.ui.internal.genericeditor.folding.DefaultFoldingReconciler" java.lang.NullPointerException: Cannot invoke "org.eclipse.jface.text.IDocument.getNumberOfLines()" because "this.document" is null
	at org.eclipse.ui.internal.genericeditor.folding.IndentFoldingStrategy.reconcile(IndentFoldingStrategy.java:234)
	at org.eclipse.ui.internal.genericeditor.folding.DefaultFoldingReconciler.process(DefaultFoldingReconciler.java:48)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundThread.run(AbstractReconciler.java:207)

```
This probably happens because `IndentFoldingStrategy.uninstall` is invoked during the reconcile run.

The issue happens only on GitHub Actions  MacOS runner (at least I never spot it when running tests locally nor while browsing the logs from  GitHub Actions Windows runner nor the Jenkins (Linux) one).

PR #229, opened 2023-06-16
[GenericEditor] An NPE hoccurs in IndentFoldingStrategy.reconcile #228

vrubezhny:
Fixes: #228

### DIFF
--- a/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/folding/IndentFoldingStrategy.java
+++ b/org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/folding/IndentFoldingStrategy.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2018 IBM Corporation and others.
+ * Copyright (c) 2009, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -223,6 +223,11 @@ public class IndentFoldingStrategy implements IReconcilingStrategy, IReconciling
 			int tabSize = 1;
 			int minimumRangeSize = 1;
 			try {
+				var thisDocument = document;
+				if (thisDocument == null) {
+					// Exit as soon as possible if uninstalled
+					return;
+				}
 
 				// Today we recompute annotation from the whole document each
 				// time.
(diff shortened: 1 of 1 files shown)


org.eclipse.ui.genericeditor/src/org/eclipse/ui/internal/genericeditor/folding/IndentFoldingStrategy.java
```diff
@@ -243,10 +248,15 @@ public void reconcile(DirtyRegion dirtyRegion, IRegion subRegion) {
...
 				int line = endLine;
 				for (line = endLine; line >= 0; line--) {
-					int lineOffset = document.getLineOffset(line);
-					String delim = document.getLineDelimiter(line);
-					int lineLength = document.getLineLength(line) - (delim != null ? delim.length() : 0);
-					String lineContent = document.get(lineOffset, lineLength);
+					thisDocument = document;
+					if (thisDocument == null) {
+						// Exit as soon as possible if uninstalled
+						return;
```
mickaelistria, line 254: Why repeating the block here? Can't we just assume that the earlier setting of thisDocument line 226 is enough?

BeckerWdf: good catch

vrubezhny: @mickaelistria This repeated block is added into a cycle to make it stop calculations as soon as possible as we don't need the results anymore.

It looks like `document` becomes `null` after a call to `uninstall` method from a different thread:
```
	public void uninstall() {
		setDocument(null);

		if (viewer != null) {
			viewer.removeProjectionListener(this);
			viewer = null;
		}

		projectionDisabled();
	}
```
So we're getting caught between  `setDocument(null)` and `projectionDisabled` calls - the check for `projectionAnnotationModel != null` in  the beginning of `reconcile` method tells me that document is already `null` while `projectionAnnotationModel` is still not. 
So we can have the same at any time during the `reconcile` method work despite the fact that `uninstall` method in fact consist of only 3 (three) method calls. And as such, the appearance of `null` document inside the `for`-loop  is quite possible after it was not `null` in the first check.  Also, if during the cycle we get `uninstall` invoked then we don't need to continue the processing cycle because we don't need its results anymore. IMHO, it's better to stop and exit it as soon as possible.

mickaelistria: > This repeated block is added into a cycle to make it stop calculations as soon as possible as we don't need the results anymore.

OK, but you don't need to repeat the assignment then.
You could even include the condition in the `for` loop about `for (line = endLine; line >= 0 && this.document != null; line--) {`

vrubezhny: Yes, you're right.

